PR #2, opened 2019-11-08
Remove namespace from manifest

MarcusNoble:
*Issue #, if available:* N/A

*Description of changes:*
Don't specify the namespace on the resources in the manifest so users can deploy to whichever namespace they choose by using `--namespace`


MarcusNoble:
Just realised it's required on the ClusterRoleBinding

Issue #25, opened 2019-11-25
Need to push git tags for at least v1.0.0 bits


jaypipes:
There are no git tags on the aws/aws-node-termination-handler repository, therefore there are no releases listed. However, there is a Docker *image* tag of v1.0.0 in dockerhub. There should be a corresponding git tag of v1.0.0 as well.

Docker image tags should really be derived from the git tag, making the git history the source of truth for all artifact versioning.

I can work on fixing up the Makefile to ensure proper git tags are used for tagging the Docker images.

bwagner5:
I took a crack at this, let me know if this is what you were thinking.

bwagner5:
v1.0.0 released via github and automation completed to automatically push to dockerhub

PR #26, opened 2019-11-25
update makefile to support versioning w/ git tags and tests

bwagner5:
*Issue #, if available:*
https://github.com/aws/aws-node-termination-handler/issues/25

*Description of changes:*
 - Added git tag versioning to makefile
 - added test targets in makefile

If this PR is accepted I'll do a git annotated tag v1.0.0 release on this commit. 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
-VERSION ?= 1.0.0-SNAPSHOT
+VERSION = $(shell git describe --tags --always --dirty)
 IMG ?= amazon/aws-node-termination-handler
-IMG_W_TAG = ${IMG}:v${VERSION}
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,6 @@
-VERSION ?= 1.0.0-SNAPSHOT
+VERSION = $(or $(shell git describe --exact-match HEAD 2> /dev/null), latest)
```
jaypipes: We shouldn't be using "latest". I'd recommend always using an image tag that refers to an immutable bunch of bits. The git SHA1 or most recent tag works well for this:

`git describe --tags --always --dirty`

bwagner5: gotcha, I updated the version generation to use `git describe --tags --always --dirty`

PR #31, opened 2019-12-02
Update Tests using KinD

bwagner5:
*Issue #, if available:*
N/A

*Description of changes:*
- Upgrade to kind v0.6.0
- Restructured repo to use `pkg` and `cmd` dirs for main repo (and with that go sub-pkgs)


go.mod
```diff
@@ -1,9 +1,10 @@
-module NodeTerminationHandler
+module github.com/aws/aws-node-termination-handler
```
jaypipes: :+1:

pkg/drainer/drainer.go
```diff
@@ -0,0 +1,98 @@
...
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package drainer
+
+import (
+	nthAppConfig "github.com/aws/aws-node-termination-handler/pkg/config"
```
jaypipes: Was "config" not appropriate to use here? Seems like "nthAppConfig" is a bit verbose.

pkg/drainer/drainer.go
```diff
@@ -0,0 +1,98 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/kubectl/pkg/drain"
+	"log"
```
jaypipes: Typically, you want to have standard imports above 3rd party (and "local") imports. Running `go fmt && go imports` on your code would automatically rearrange these imports accordingly.

https://godoc.org/golang.org/x/tools/cmd/goimports

Cool to leave this as-is though and I can push a new PR that adds a `go fmt/imports` check test.

bwagner5: ah yeah, I had switched my formatter in vs code because it was removing packages when I was breaking stuff out for the reorg.

bwagner5: I'll break up this PR into better chunks :)

PR #32, opened 2019-12-02
update kind to v0.6.0 and move ec2-meta-data-proxy to test root

bwagner5:
*Issue #, if available:*
N/A

*Description of changes:*
- Updated KinD to v0.6.0 
- Moved the test metadata into test root so that it can be used across future tests


test/ec2-meta-data-proxy/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/aws/aws-node-termination-handler/ec2-meta-data-proxy
```
jaypipes: Let's steer clear of Go submodules in the repo. multi-module repos are an extreme pain in the you-know-what. I've learned this from painful experience recently:

https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository
https://github.com/golang/go/issues/35878

In short, if we used multiple Go modules in a single repo, we'd need to remember to apply a semver "prefixed" Git tag every time we issued a release, one for each Go submodule. It's a recipe for disaster and we can do just fine with a single semver Git-tagged repo with a single root Go.mod.

bwagner5:
yeah that's not a bad idea, I think I like `ec2-metadata-test-proxy`, I'll update

Issue #29, opened 2019-11-27
Add configurable delay to interruption handling


mattrandallbecker:
Currently the daemonset immediately cordons and drains a node upon ITN detection.

We should allow users to define a period to wait before initializing cordon and drain such that they can still perform work in that window.  e.g. if you can successfully cordon and drain in 30 seconds, delay the process 90 seconds.

bwagner5:
functionality to set a configurable node-termination-grace-period has been merged, probably need some better docs

bwagner5:
I'll open a general issue for better docs on configurable values

PR #35, opened 2019-12-03
node-termination-grace-period configuration option added

bwagner5:
*Issue #, if available:*
https://github.com/aws/aws-node-termination-handler/issues/29

*Description of changes:*
This change adds a configurable nodeTerminationGracePeriod which specifies how long a node is given to drain. This information is used within the drain logic to respond to scheduled drainable events (like spot interruption termination notice) so that every second of compute can be squeezed out of the node before a drain is initiated. The default value is 2 minutes which is the default spot ITN period. 

This also changes the configurable grace-period to pod-termination-grace-period to better differentiate the two configurable grace-periods. This code will be released in the v2.0.0 release since it breaks the backwards compatible configuration. 

If the new config changes are approved, I'll stage a PR to update the eks helm repo. 

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -23,32 +23,38 @@ import (
 
 const (
 	// EC2 Instance Metadata is configurable mainly for testing purposes
-	instanceMetadataURLConfigKey       = "INSTANCE_METADATA_URL"
-	defaultInstanceMetadataURL         = "http://169.254.169.254"
-	dryRunConfigKey                    = "DRY_RUN"
-	nodeNameConfigKey                  = "NODE_NAME"
-	kubernetesServiceHostConfigKey     = "KUBERNETES_SERVICE_HOST"
-	kubernetesServicePortConfigKey     = "KUBERNETES_SERVICE_PORT"
-	deleteLocalDataConfigKey           = "DELETE_LOCAL_DATA"
-	ignoreDaemonSetsConfigKey          = "IGNORE_DAEMON_SETS"
-	podTerminationGracePeriodConfigKey = "GRACE_PERIOD"
+	instanceMetadataURLConfigKey        = "INSTANCE_METADATA_URL"
+	defaultInstanceMetadataURL          = "http://169.254.169.254"
+	dryRunConfigKey                     = "DRY_RUN"
+	nodeNameConfigKey                   = "NODE_NAME"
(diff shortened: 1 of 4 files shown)


pkg/config/config.go
```diff
@@ -56,7 +59,8 @@ func ParseCliArgs() Config {
 	flag.BoolVar(&config.DeleteLocalData, "delete-local-data", getBoolEnv(deleteLocalDataConfigKey, true), "If true, do not drain pods that are using local node storage in emptyDir")
 	flag.StringVar(&config.KubernetesServiceHost, "kubernetes-service-host", getEnv(kubernetesServiceHostConfigKey, ""), "[ADVANCED] The k8s service host to send api calls to.")
 	flag.StringVar(&config.KubernetesServicePort, "kubernetes-service-port", getEnv(kubernetesServicePortConfigKey, ""), "[ADVANCED] The k8s service port to send api calls to.")
-	flag.IntVar(&config.PodTerminationGracePeriod, "grace-period", getIntEnv(podTerminationGracePeriodConfigKey, -1), "Period of time in seconds given to each pod to terminate gracefully. If negative, the default value specified in the pod will be used.")
```
jaypipes: Let's be good release stewards and provide this functionality in the 1.x release series. Simply deprecate the original `--grace-period` CLI option and note its rename to `--pod-termination-grace-period`.

bwagner5: here's some simple testing:  https://gist.github.com/bwagner5/b18ddb5576b6d6dad13b5bd13dc1905d

pkg/ec2metadata/ec2metadata.go
```diff
@@ -61,9 +61,10 @@ func CheckForSpotInterruptionNotice(metadataURL string) bool {
 		log.Fatalln("Could not parse time from metadata json", err.Error())
 	}
 	timeUntilInterruption := time.Now().Sub(interruptionTime)
-	if timeUntilInterruption <= (time.Duration(120) * time.Second) {
+	if timeUntilInterruption <= (time.Duration(nodeTerminationGracePeriod) * time.Second) {
 		return true
 	}
+	log.Printf("Termination notice received, but the time until interruption is %d and the node is configured to drain for %d sec. Waiting to drain...", timeUntilInterruption, nodeTerminationGracePeriod)
```
jaypipes: The format string should have a "\n" at the end of it.

bwagner5: updated

PR #37, opened 2019-12-11
Added statement to print Interruption notice metadata for keeping track of Interrupted instances

opuzaman:
*Issue #, if available:*
None
*Description of changes:*
Added statement to print Interruption notice metadata for keeping track of Interrupted instances. My manager wanted to see which instance was terminated and more details about the notice as a proof of spot interruptions. 

### DIFF
--- a/pkg/ec2metadata/ec2metadata.go
+++ b/pkg/ec2metadata/ec2metadata.go
@@ -57,7 +57,7 @@ func CheckForSpotInterruptionNotice(metadataURL string, nodeTerminationGracePeri
 	var instanceAction InstanceAction
 
 	json.NewDecoder(resp.Body).Decode(&instanceAction)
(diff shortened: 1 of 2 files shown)


pkg/ec2metadata/ec2metadata.go
```diff
@@ -55,7 +55,9 @@ func CheckForSpotInterruptionNotice(metadataURL string, nodeTerminationGracePeri
 		return false
 	}
 	var instanceAction InstanceAction
+
 	json.NewDecoder(resp.Body).Decode(&instanceAction)
+	log.Printf("Spot Instance Interruption Warning received. Instance metadata: %v", instanceAction)
```
bwagner5: this needs a newline character added. 

Also, since we're printing the struct, we probably want `%+v` to print field names

PR #42, opened 2020-01-02
use channels to coordinate drainable events

bwagner5:
*Issue #, if available:*
N/A

*Description of changes:*

- Refactor the event handling logic to be have more generic event handling that does not need to be touched when adding an additional drainable event.

- This PR adds a go routine which blocks on a go channel to receive drain events from monitor loop sources. The only monitor loop involved right now is the spot ITN loop. As more events types are introduced, the developer simply needs to write a monitor loop that uses the function interface and provide the function to the list of monitor functions.

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -22,41 +22,59 @@ import (
 
 	"github.com/aws/aws-node-termination-handler/pkg/config"
 	"github.com/aws/aws-node-termination-handler/pkg/drainer"
-	"github.com/aws/aws-node-termination-handler/pkg/ec2metadata"
+	"github.com/aws/aws-node-termination-handler/pkg/drainevent"
 )
 
 func main() {
-	setupSignalHandler()
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, syscall.SIGTERM)
+	defer signal.Stop(signalChan)
+
 	nthConfig := config.ParseCliArgs()
 	drainer.InitDrainer(nthConfig)
+	drainEventStore := drainevent.NewStore(&nthConfig)
 
-	log.Println("Kubernetes Spot Node Termination Handler has started successfully!")
-	waitForTermination(nthConfig)
-	drainer.Drain(nthConfig)
-	log.Printf("Node %q successfully drained.\n", nthConfig.NodeName)
+	log.Println("Kubernetes AWS Node Termination Handler has started successfully!")
+	drainChan := make(chan drainevent.DrainEvent)
+	defer close(drainChan)
+	monitoringFns := []func(chan<- drainevent.DrainEvent, config.Config){
+		drainevent.MonitorForSpotITNEvents,
+	}
+	for _, fn := range monitoringFns {
+		go fn(drainChan, nthConfig)
+	}
 
-	// Sleep to prevent process from restarting.
-	// The node should be terminated after configured drain time
-	time.Sleep(time.Duration(nthConfig.NodeTerminationGracePeriod) * time.Second)
-}
+	go watchForDrainEvents(drainChan, drainEventStore, nthConfig)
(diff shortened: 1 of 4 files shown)


cmd/node-termination-handler.go
```diff
@@ -22,33 +22,47 @@ import (
...
 	"github.com/aws/aws-node-termination-handler/pkg/drainer"
-	"github.com/aws/aws-node-termination-handler/pkg/ec2metadata"
+	"github.com/aws/aws-node-termination-handler/pkg/drainevent"
 )
 
 func main() {
 	setupSignalHandler()
 	nthConfig := config.ParseCliArgs()
 	drainer.InitDrainer(nthConfig)
+	drainEventStore := drainevent.Store{
```
jaypipes: suggestion: a more idiomatic way of doing this would be to have a `drainevent.NewStore()` function that populated a constructed `drainevent.Store` struct. In that `NewStore` function, you could ensure that, for instance, `Store.drainEvents` is `make(map[string]*DrainEvent)` so you don't have to have a conditional like this:

```go
if s.drainEvents == nil {
```

in methods like `drainEventStore.AddDrainEvent()` below.

pkg/drainevent/spot-itn-event.go
```diff
@@ -0,0 +1,54 @@
...
+			log.Println("Sending drain event to the drain channel")
+			drainChan <- *drainEvent
+			// cool down for the system to respond to the drain
+			time.Sleep(120 * time.Minute)
+		}
+	}
+}
+
+// CheckForSpotInterruptionNotice Checks EC2 instance metadata for a spot interruption termination notice
+func CheckForSpotInterruptionNotice(metadataURL string) *DrainEvent {
```
jaypipes: this function doesn't need to be exported and should be made private.

cmd/node-termination-handler.go
```diff
@@ -22,33 +22,47 @@ import (
...
+		NthConfig: &nthConfig,
+	}
+
+	log.Println("Kubernetes AWS Node Termination Handler has started successfully!")
+	drainChan := make(chan drainevent.DrainEvent)
```
jaypipes: technically you should `close()` this channel. potentially put the chan definition in the module scope and `close()` it in your signal handler.

pkg/drainevent/drain-event.go
```diff
@@ -0,0 +1,93 @@
...
+
+	drained bool
+}
+
+// Store is a the drain event store data structure
+type Store struct {
+	NthConfig *config.Config
+
+	drainEventStore map[string]*DrainEvent
+	mutex           sync.RWMutex
```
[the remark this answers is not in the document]
bwagner5: Fixed, I like the clean-ness

cmd/node-termination-handler.go
bwagner5: Cool, yep, I like it. Fixed

cmd/node-termination-handler.go
bwagner5: I thought about this with mrbecker a bit and there really isn't a need for the signal handler we have now anyways... so I refactored the code to exit normally and then just used defers.

jaypipes: Cool with me! :)

pkg/drainevent/drain-event.go
```diff
@@ -0,0 +1,97 @@
...
+// TimeUntilEvent returns the duration until the event start time
+func (e *DrainEvent) TimeUntilEvent() time.Duration {
+	return e.StartTime.Sub(time.Now())
+}
+
+// NewStore Create a new drain event store
+func NewStore(nthConfig *config.Config) *Store {
+	return &Store{
+		NthConfig:       nthConfig,
+		drainEventStore: make(map[string]*DrainEvent),
```
jaypipes: boom.

PR #43, opened 2020-01-02
add ec2-metadata-test-proxy helm chart

bwagner5:
*Issue #, if available:*

*Description of changes:*

- Add an ec2-metadata-test-proxy helm chart so we can make the tests a bit smaller by getting rid of kustomize heredocs


config/helm/ec2-metadata-test-proxy/templates/daemonset.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if .Values.ec2MetadataTestProxy.create -}}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: ec2-metadata-test-proxy
```
jillmon: This should be a variable name.

PR #48, opened 2020-01-07
setup release scripts

bwagner5:
*Issue #, if available:*

*Description of changes:*
 - Build multiple platform binaries with the docker image 
 - Upload built platform binaries to github release assets upon tag push
 - Sync github readme with dockerhub full_description

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,5 +27,11 @@ matrix:
       script: make helm-sync-test
       env: HELM_SYNC_TEST=true
     - stage: Deploy
+      if: type = push AND env(DOCKER_USERNAME) IS present
+      script: make sync-readme-to-dockerhub
+    - stage: Deploy
       if: type = push AND tag =~ /^v\d+\.\d+(\.\d+)?(-\S*)?$/ AND env(DOCKER_USERNAME) IS present
-      script: make docker-build && make docker-push
+      script: make docker-build-and-push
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -26,6 +26,14 @@ matrix:
     - stage: Test
       script: make helm-sync-test
       env: HELM_SYNC_TEST=true
+    - stage: Deploy
+      if: type = push AND env(DOCKER_USERNAME) IS present
+      script: make sync-readme-to-dockerhub
```
mattrandallbecker: should we be doing this on every push? can we detect if there was actually a change to the readme before replacing?

scripts/upload-binaries-to-github
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
mattrandallbecker: this script doesn't seem to be referenced in the makefile

.travis.yml
```diff
@@ -26,6 +26,14 @@ matrix:
...
     - stage: Deploy
       if: type = push AND tag =~ /^v\d+\.\d+(\.\d+)?(-\S*)?$/ AND env(DOCKER_USERNAME) IS present
       script: make docker-build && make docker-push
```
mattrandallbecker: maybe alias to a single command for consistency?

PR #58, opened 2020-01-14
Added missing namespace references to helm chart

gavinbunney:
When rendering out the helm chart to template files, the generated manifests don't have namespaces for all the namespaced resources. This fix adds the missing `namespace` to the `DaemonSet` and `ServiceAccount` resources.


jillmon:
Everything looks good, but we have a few other changes in the pipeline. We're going to wait to merge yours till the other two are completed. That way we can cut a single PR to eks-charts instead of three different ones.

Issue #30, opened 2019-11-27
Handle instance retirement events


johanneswuerbach:
From time to time AWS is retiring EC2 instances due to underlying hardware or software failures https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/monitoring-instances-status-check_sched.html#types-of-scheduled-events

I would be awesome if the aws-node-termination-handler could watch the instance meta data service to react automatically to those events https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/monitoring-instances-status-check_sched.html#viewing_scheduled_events and drain the affected instance.

Would this be something in scope of this project?

mattrandallbecker:
Yep, definitely within the wheelhouse of this project.  Thanks for the suggestion!

mattrandallbecker:
resolved with 1.2.0 release

https://github.com/aws/aws-node-termination-handler/releases/tag/v1.2.0

keep in mind this feature is off by default for this release but can be enabled via the helm flag outlined in the readme

PR #60, opened 2020-01-17
implement scheduled maintenance events

bwagner5:
Issue #, if available:

## Description of changes:
This change implements scheduled maintenance event monitoring and subsequent cordon and draining actions. 

## Testing

e2e tests have also been added to test the new maintenance event functionality. 

## Helm

Helm changes will be sync'd with eks-charts after the series of PRs related to scheduled maintenance events have been merged. 

---


pkg/drainer/drainer.go
```diff
@@ -28,50 +28,59 @@ import (
 
 var drainHelper *drain.Helper
 var node *corev1.Node
+var nthConfig config.Config
```
jaypipes: As a best practice, it's best to have as few global state variables as possible. Your code is more extensible (and testable) when you pass the config object to functions like `getDrainHelper()`. You can test the configuration scenarios more easily (in parallel) and thoroughly when you can construct new config objects and pass those into functions instead of relying on global state.

pkg/drainer/drainer.go
```diff
@@ -28,50 +28,59 @@ import (
...
 
-//InitDrainer will ensure the drainer has all the resources necessary to complete a drain
-func InitDrainer(nthConfig config.Config) {
+// InitDrainer will ensure the drainer has all the resources necessary to complete a drain.
+// This function must be called before any function in the drainer package.
+func InitDrainer(inputNthConfig config.Config) {
+	nthConfig = inputNthConfig
```
jaypipes: It's particularly odd seeing this here, where you're copy-constructing a global-scoped variable to the value of a parameter that you pass to the outer function...

pkg/drainer/drainer.go
```diff
@@ -28,50 +28,59 @@ import (
...
 	var err error
-	if !nthConfig.DryRun {
-		node, err = drainHelper.Client.CoreV1().Nodes().Get(nthConfig.NodeName, metav1.GetOptions{})
-		if err != nil {
-			log.Fatalf("Couldn't get node %q: %s\n", nthConfig.NodeName, err.Error())
-		}
-		log.Printf("Successufully retrieved node: %s", node.Name)
+	node, err = FetchNode()
+	if err != nil {
+		log.Fatalf("Couldn't get node %q: %s\n", nthConfig.NodeName, err.Error())
```
jaypipes: FYI, you don't need to use `Error()` :) You can just pass err to the formatting function and the `%s` marker will be interpolated (the error's `String()` method is called. Alternately, you can use the `%w` interpolation marker for Go1.13+ to bring in any wrapped error details.

pkg/drainer/drainer.go
```diff
@@ -28,50 +28,59 @@ import (
...
-			log.Fatalf("Couldn't cordon node %q: %s\n", node, err.Error())
-		}
+		log.Printf("Node %s would have been cordoned and drained, but dry-run flag was set\n", nthConfig.NodeName)
+		return
 	}
+	cordonNode()
+	// Delete all pods on the node
+	err = drain.RunNodeDrain(drainHelper, nthConfig.NodeName)
+	if err != nil {
+		log.Fatalln(err.Error())
```
jaypipes: Just something to do in a followup PR, not in this one...

Generally, it's best practice to never call `panic()` or the `FatalX` counterparts of the log package outside of the `main()` function. Instead, you want to return the error as-is (or wrapped using the pkg/errors `Wrap()` function) to the top-level entrypoint and have that entrypoint decide whether to segfault (panic) or print out the error to a log file or something and continue running.

See this excellent article from Dave Cheney about this topic: https://dave.cheney.net/tag/error-handling

test/ec2-metadata-test-proxy/cmd/ec2-metadata-test-proxy.go
```diff
@@ -73,6 +90,18 @@ func getListenAddress() string {
 	return ":" + port
 }
 
+func initEnabledPaths() {
+	pathsEnabled = make(map[string]bool)
+	if strings.ToUpper(getEnv(spotInstanceActionFlag, "")) == "TRUE" {
```
jaypipes: https://golang.org/pkg/strconv/#ParseBool :)

test/ec2-metadata-test-proxy/cmd/ec2-metadata-test-proxy.go
```diff
@@ -36,9 +37,25 @@ const (
...
 	interruptionNoticeDelayDefault   = "0"
+	scheduledActionDateFormat        = "02 Jan 2006 15:04:05 GMT"
+	spotInstanceActionFlag           = "ENABLE_SPOT_ITN"
+	spotInstanceActionPath           = "/latest/meta-data/spot/instance-action"
+	scheduledMaintenanceEventFlag    = "ENABLE_SCHEDULED_MAINTENANCE_EVENTS"
+	scheduledMaintenanceEventPath    = "/latest/meta-data/events/maintenance/scheduled"
 )
 
 var startTime int64 = time.Now().Unix()
+var pathsEnabled map[string]bool
```
jaypipes: same comment about generally avoiding global state...

test/ec2-metadata-test-proxy/cmd/ec2-metadata-test-proxy.go
```diff
@@ -73,6 +90,18 @@ func getListenAddress() string {
 	return ":" + port
 }
 
+func initEnabledPaths() {
```
jaypipes: Not sure this is useful to cache this kind of thing. Doing a `strconv.ParseBool` (see below) on an environs variable is not a costly operation that you need to worry about, so I'd recommend getting rid of the global state and simplifying things with just a simple function like this:

```go
func spotInstanceActionEnabled() bool {
    val, _ := strconv.ParseBool(getEnv(spotInstanceActionPath, ""))
    return val 
}
```

pkg/drainer/drainer.go
bwagner5: Refactored drainer file

pkg/drainer/drainer.go
bwagner5: yep, agreed! refactored drainer file

pkg/drainer/drainer.go
bwagner5: ah good to know, I removed the Error() in drainer. There's probably more places that this should be simplified, but I'll get to that in a cleanup PR.

pkg/drainer/drainer.go
bwagner5: I went ahead and removed all panics from drainer in this PR since I was refactoring drainer anyways.  I think there are some more places this needs to be taken care of, but again, I'll include those changes in a cleanup PR after the sequence of scheduled maintenance events PRs get merged.

test/ec2-metadata-test-proxy/cmd/ec2-metadata-test-proxy.go
bwagner5: function removed and used ParseBool instead of poor mans parsing in the new function :)

test/ec2-metadata-test-proxy/cmd/ec2-metadata-test-proxy.go
bwagner5: removed

pkg/drainer/drainer.go
```diff
@@ -26,74 +26,97 @@ import (
...
-		Out:                 os.Stdout,
-		ErrOut:              os.Stderr,
+// Cordon will add a NoSchedule on the node
+func (d Drainer) cordonNode() error {
+	node, err := d.FetchNode()
+	if err != nil {
+		log.Println("There was an error which fetching the kubernetes node: ", err)
+		return err
+	}
+	err = drain.RunCordonOrUncordon(d.drainHelper, node, true)
```
nithu0115: Before calling cordon, can we add an extra check to see if node is already marked unschedulable ? 

```
if node.Spec.Unschedulable {
		log.Infof("node '%s' is already unschedulable.\n", d.node)
		return nil
	}

Comment (bwagner5) [reply pkg/drainer/drainer.go]: I'm not sure this is necessary unless it just serves as an informational message. It doesn't really do any harm in calling cordon again if the node is already cordoned. Kubernetes is just going to converge to the desired state. If it's already cordoned, there will just be less work to do to get there.

Comment (jaypipes) [reply pkg/drainer/drainer.go]: Yes, agreed with @bwagner5 here. The whole cordon process really is just tainting the node as unscheduleable, and it's an idempotent process, so I don't believe another check here is necessary.

File: config/helm/ec2-metadata-test-proxy/values.yaml [hunk 9]
```diff
@@ -4,6 +4,8 @@ ec2MetadataTestProxy:
     interruptionNoticeDelay: 15
     port: 1338
     label: ec2-metadata-test-proxy
+    enableScheduledMaintenanceEvents: false
```
jaypipes: meh, it's just the test proxy :) might as well default this to true, since it's not like we're worried about backwards compat or anything.

pkg/drainer/drainer.go
```diff
@@ -26,74 +26,97 @@ import (
...
-	}
-	node = &corev1.Node{}
-	var err error
-	if !nthConfig.DryRun {
-		node, err = drainHelper.Client.CoreV1().Nodes().Get(nthConfig.NodeName, metav1.GetOptions{})
-		if err != nil {
-			log.Fatalf("Couldn't get node %q: %s\n", nthConfig.NodeName, err.Error())
-		}
-		log.Printf("Successufully retrieved node: %s", node.Name)
+// InitDrainer will ensure the drainer has all the resources necessary to complete a drain.
```
jaypipes: Left the old doc comment in here...

pkg/drainer/drainer.go
```diff
@@ -26,74 +26,97 @@ import (
...
+// This function must be called before any function in the drainer package.
+func New(nthConfig config.Config) (*Drainer, error) {
+	drainHelper, err := getDrainHelper(nthConfig)
+	if err != nil {
+		log.Println("Unable to construct a drainer because a kubernetes drainHelper could not be built: ", err)
+		return &Drainer{}, err
```
jaypipes: best practice would be to return `nil, err` here, not a pointer to an actual Drainer struct.

pkg/drainer/drainer.go
```diff
@@ -26,74 +26,97 @@ import (
...
-	} else {
-		// Delete all pods on the node
-		err = drain.RunNodeDrain(drainHelper, nthConfig.NodeName)
-		if err != nil {
-			log.Fatalln(err.Error())
-		}
+	clusterConfig, err := rest.InClusterConfig()
+	if err != nil {
+		log.Println("Failed to create in-cluster config: ", err)
+		return &drain.Helper{}, err
```
jaypipes: `return nil, err`

pkg/drainer/drainer.go
```diff
@@ -26,74 +26,97 @@ import (
...
 	}
+
+	// creates the clientset
+	clientset, err := kubernetes.NewForConfig(clusterConfig)
+	if err != nil {
+		log.Println("Failed to create kubernetes clientset: ", err)
+		return &drain.Helper{}, err
```
jaypipes: `return nil, err`

PR #61, opened 2020-01-20
implement cancellation of drain events

bwagner5:
Issue #, if available:

# Drain Event Cancellation Handling

## Description of changes:

This PR implements a cancel drain event channel and handler function. The new cancel drain channel is passed to all monitor functions. The Monitor functions are responsible for the cancellation logic for each specific case.  The Spot ITN monitoring function accepts the cancel channel as an argument but ignores it since there is never a case where we should cancel a spot ITN drain event. Scheduled maintenance events are able to be cancelled, either by the AWS user or by EC2. The monitor loop will watch for status changes in the metadata response for maintenance events and propagate those to the cancel channel. 

Once a cancel event is received, the root cancel handler will query the drain event store data structure to see if it should uncordon the node.

## Additional Changes:

 - The drainer package has been renamed to node since it now has more functionality than just draining and cordoning nodes. It can now uncordon nodes.

 ## Tests:

 - e2e test for cancelling a scheduled maintenance event. 

---

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -21,9 +21,9 @@ import (
 	"time"
 
 	"github.com/aws/aws-node-termination-handler/pkg/config"
-	"github.com/aws/aws-node-termination-handler/pkg/drainer"
 	"github.com/aws/aws-node-termination-handler/pkg/drainevent"
 	"github.com/aws/aws-node-termination-handler/pkg/draineventstore"
+	"github.com/aws/aws-node-termination-handler/pkg/node"
 	"github.com/aws/aws-node-termination-handler/pkg/webhook"
 )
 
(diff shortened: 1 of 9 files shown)


cmd/node-termination-handler.go
```diff
@@ -33,26 +33,31 @@ func main() {
...
 	}
 	drainEventStore := draineventstore.New(&nthConfig)
 
 	drainChan := make(chan drainevent.DrainEvent)
 	defer close(drainChan)
+	cancelChan := make(chan drainevent.DrainEvent)
+	defer close(cancelChan)
 
-	monitoringFns := []func(chan<- drainevent.DrainEvent, config.Config){}
+	monitoringFns := []func(chan<- drainevent.DrainEvent, chan<- drainevent.DrainEvent, config.Config){}
```
jaypipes: consider type-aliasing the callback definition to clean this up a bit:

```go
type MonitorFunc func(chan<- drainevent.DrainEvent, chan<- drainevent.DrainEvent, config.Config)
...
monitoringFns := []MonitorFunc{
    draineventstore.MonitorForSpotITNEvents,
    draineventstore.MonitorForScheduledEvents,
}
```

pkg/draineventstore/spot-itn-event.go
```diff
@@ -17,15 +17,15 @@ const (
...
+func MonitorForSpotITNEvents(drainChan chan<- drainevent.DrainEvent, cancelChan chan<- drainevent.DrainEvent, nthConfig config.Config) {
 	log.Println("Started monitoring for spot ITN events")
 	for range time.Tick(time.Second * 2) {
 		drainEvent := checkForSpotInterruptionNotice(nthConfig.MetadataURL)
 		if drainEvent.Kind == SpotITNKind {
 			log.Println("Sending drain event to the drain channel")
 			drainChan <- *drainEvent
 			// cool down for the system to respond to the drain
-			time.Sleep(120 * time.Minute)
+			time.Sleep(120 * time.Second)
```
jaypipes: hmmm, oops :) should have caught this in the prior PR!

pkg/node/node.go
```diff
@@ -107,16 +107,35 @@ func (d Drainer) Drain() error {
...
 		log.Printf("Couldn't cordon node %q: %v\n", node, err)
 		return err
 	}
 	return nil
 }
+
+// UncordonNode will remove the NoSchedule on the node
+func (n Node) UncordonNode() error {
+	if n.nthConfig.DryRun {
+		log.Printf("Node %s would have been uncordoned, but dry-run flag was set", n.nthConfig.NodeName)
```
jaypipes: Here and below... remember that conversation about just returning a wrapped error to the top-most caller and letting the top-most caller decide whether to print.

test/e2e/maintenance-event-cancellation-test
```diff
@@ -0,0 +1,97 @@
...
+        echo "✅ Verified the worker node was uncordoned!"
+        if [[ $(kubectl get deployments regular-pod-test -o=jsonpath='{.status.unavailableReplicas}') -eq 0 ]]; then
+            echo "✅ Verified the regular-pod-test pod was rescheduled"
+            echo "✅ Scheduled Maintenance Event Cancellation Test Passed $CLUSTER_NAME! ✅"
+            exit 0
+        fi
+    fi
+    sleep $TAINT_CHECK_SLEEP
+done
+
```
jaypipes: good tests, Brandon. ++

pkg/node/node.go
bwagner5: hmm would you recommend treating the dry-run case as an error with a top-level print?  I was printing here as the actual functionality of `UncordonNode` doing the correct thing when the dry-run flag was enabled.

jaypipes: Apologies, not the dry-run ones :) Just the real errors.

PR #65, opened 2020-01-23
[1] - Label action handling for uncordoning after reboot

bwagner5:
Issue #, if available:

# Maintenance Events Uncordon After Reboot

## Description of changes:
This change supports labeling the current node with an action tag and action-time tag in order to support uncordoning the node after a maintenance event requiring a reboot.  This automates the process of cordoning nodes temporarily going down for maintenance and then removing the cordon when the maintenance has been completed. 

 ## More Details:

 - If a `system-reboot` maintenance event is received, the scheduled-maintenance-event monitor loop will add a new `PreDrainTask` to the `drainevent` which is a function that executes right before the node is cordoned and drained. In the `system-reboot` case, the function checks if the node is already cordoned (meaning an out-of-band process cordoned the node), in which case the function will exit. If the node is not already cordoned, the `PreDrainTask` will apply a label to the node `aws-node-termination-handler/action=UncordonAfterReboot` and `aws-node-termination-handler/action-time=<seconds_since_epoch>`.  When the aws-node-termination-handler starts up, it will check for these labels on the node. If it finds them, it will check the current system uptime and determine if a reboot has taken place. It will then uncordon the node and remove the labels. 

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -27,15 +27,25 @@ import (
 	"github.com/aws/aws-node-termination-handler/pkg/webhook"
 )
 
+type monitorFunc func(chan<- drainevent.DrainEvent, chan<- drainevent.DrainEvent, config.Config)
+
 func main() {
 	signalChan := make(chan os.Signal, 1)
 	signal.Notify(signalChan, syscall.SIGTERM)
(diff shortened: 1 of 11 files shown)


bwagner5:
Helm PR is here: https://github.com/aws/eks-charts/pull/62

PR #73, opened 2020-01-27
monitor functions return errs and loop in main

bwagner5:
Issue #, if available:
N/A

Description of changes:

This is a cleanup and slight restructure of the monitor loops for drain events. This pulls out the for loops into the main cmd so that the monitor call can propagate errors up.  It also more clearly shows the monitor looping interval and will make adding unit tests easier. 

- Removed logging an error on IMDS 404 for SPOT ITN since this is a proper response from that API. If no spot interruption exists, a 404 is sent. 

---

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -27,7 +27,7 @@ import (
 	"github.com/aws/aws-node-termination-handler/pkg/webhook"
 )
 
-type monitorFunc func(chan<- drainevent.DrainEvent, chan<- drainevent.DrainEvent, config.Config)
+type monitorFunc func(chan<- drainevent.DrainEvent, chan<- drainevent.DrainEvent, config.Config) error
 
 func main() {
 	signalChan := make(chan os.Signal, 1)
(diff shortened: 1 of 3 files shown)


cmd/node-termination-handler.go
```diff
@@ -58,16 +58,24 @@ func main() {
 	cancelChan := make(chan drainevent.DrainEvent)
 	defer close(cancelChan)
 
-	monitoringFns := []monitorFunc{}
+	monitoringFns := map[string]monitorFunc{}
 	if nthConfig.EnableSpotInterruptionDraining {
-		monitoringFns = append(monitoringFns, drainevent.MonitorForSpotITNEvents)
+		monitoringFns["Spot ITN"] = drainevent.MonitorForSpotITNEvents
```
jaypipes: consider making these strings constants.

pkg/drainevent/scheduled-event.go
```diff
@@ -20,54 +20,62 @@ const (
...
 
-// CheckForScheduledEvents Checks EC2 instance metadata for a scheduled event requiring a node drain
-func CheckForScheduledEvents(metadataURL string) []DrainEvent {
-	events := make([]DrainEvent, 0)
+// checkForScheduledEvents Checks EC2 instance metadata for a scheduled event requiring a node drain
+func checkForScheduledEvents(metadataURL string) ([]DrainEvent, error) {
 	resp, err := ec2metadata.RequestMetadata(metadataURL, ec2metadata.ScheduledEventPath)
 	if err != nil {
-		log.Fatalf("Unable to parse metadata response: %s", err.Error())
+		return nil, fmt.Errorf("Unable to parse metadata response: %w", err)
```
jaypipes: note that `%w` is go 1.13+. Is NTH only built on 1.13 and up? If so, cool. if not, well, prolly want to use `%v` or `%s` here.

jaypipes: Alternately, you could always return an `errors.Wrapf()` here instead?

https://godoc.org/github.com/pkg/errors#Wrapf

pkg/drainevent/scheduled-event.go
```diff
@@ -91,11 +99,3 @@ func isStateCancelledOrCompleted(state string) bool {
...
 		state == scheduledEventStateCompleted
 }
-
-func parseScheduledEventTime(inputTime string) time.Time {
-	scheduledTime, err := time.Parse(scheduledEventDateFormat, inputTime)
-	if err != nil {
-		log.Fatalln("Could not parse time from scheduled event metadata json", err.Error())
-	}
-	return scheduledTime
-}
```
jaypipes: :+1: getting rid of this in favour of inlining the checks above.

pkg/drainevent/scheduled-event.go
bwagner5: NTH is only built on 1.13 (the binary distributions and docker images). I suppose the only concern is if someone is building from source and making their own distribution images. `%w` is used elsewhere in the codebase, so I'll go ahead and merge this one. We can chat if this is a big issue or not.

PR #76, opened 2020-01-29
Run unit tests in travis build

jillmon:
Issue #, if available:

Should run the unit tests....

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: minimal
 services:
 - docker
 
-env: 
+env:
 - KUBE_VERSION=1.17
 - KUBE_VERSION=1.16
 - KUBE_VERSION=1.15
@@ -16,7 +16,13 @@ env:
 script: test/k8s-local-cluster-test/run-test -v $KUBE_VERSION
 
 matrix:
-  include: 
+  include:
+    - stage: Test
+      language: go
+      go: "1.13.x"
+      script: make unit-test
+      env: GO111MODULE=on
+      after_success: bash <(curl -s https://codecov.io/bash)
     - stage: Test
       script: test/go-report-card-test/run-report-card-test.sh
       env: GO_REPORT_CARD=true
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -16,7 +16,16 @@ env:
...
 
 matrix:
-  include: 
+  include:
+    - stage: Test
+      language: go
+      go:
+        - 1.13
+      before_script:
+        - go get -t -v ./...
```
bwagner5: I don't think you need this since we're using go modules

.travis.yml
```diff
@@ -16,7 +16,16 @@ env:
...
+      script: go test ./... -coverprofile=coverage.txt -covermode=atomic
```
bwagner5: this should ideally go in the Makefile so we can just run `make unit` and then add as a dependent target in `make test`. The coverage.txt output should go to the build dir so that we don't accidentally check it in. The codecov bash script looks like it checks subdirs from the root so it should still find it or you can pass in the abs path to the codecov script.

.travis.yml
jaypipes: Just set the `GO111MODULE: "on"` environment variable and, yes, you don't need the above.

.travis.yml
jaypipes: Yes, that's a best practice, Brandon. Have a `make unit-test` target (or similar) and run `make unit-test` from the CI job.

Example from VPC CNI:

https://github.com/aws/amazon-vpc-cni-k8s/blob/8a59da0e3e29bc06a9d419f212cde0d4897cd6c6/Makefile#L57-L58

https://github.com/aws/amazon-vpc-cni-k8s/blob/8a59da0e3e29bc06a9d419f212cde0d4897cd6c6/.circleci/config.yml#L18

Makefile
```diff
@@ -68,4 +68,7 @@ docker-build-and-push: docker-build docker-push
 
 test: e2e-test compatibility-test license-test go-report-card-test helm-sync-test
 
+unit-test: 
+	go test ./... -v -coverprofile=build/coverage.txt -covermode=atomic
```
bwagner5: use the `${BUILD_DIR_PATH}/coverage.txt`

bwagner5: also use `go test ${MAKEFILE_PATH}/...` so that we can reference the Makefile from outside the root like this: `make -f ../../Makefile unit-test`

PR #80, opened 2020-01-30
Update config to return err & add unit tests

jillmon:
Description of changes:
Rewrote config.go to return errors instead of panicking in place. As a result, flags and their default values are all declared in an object of type map[string]map[string]interface{}. Each flag is then evaluated for validity and the resulting values are stored in a map[string]interface{} which is eventually json marshaled into the nthConfig. The new changes were necessary to help consolidate error checking and avoid repeatedly checking if error else do something. 

Changes fully unit tested: 
```
ok  	github.com/aws/aws-node-termination-handler/pkg/config	0.027s	coverage: 98.6% of statements
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -68,84 +68,90 @@ type Config struct {
 	WebhookTemplate                string `json:"webhook-template"`
 }
 
-var flagData = map[string]map[string]interface{}{
-	"delete-local-data": map[string]interface{}{
-		"key":      deleteLocalDataConfigKey,
-		"defValue": true,
-		"usage":    "If true, do not drain pods that are using local node storage in emptyDir",
+type flagData struct {
+	Key      string
+	DefValue interface{}
+	Usage    string
+}
+
+var flags = map[string]flagData{
+	"delete-local-data": flagData{
+		Key:      deleteLocalDataConfigKey,
+		DefValue: true,
+		Usage:    "If true, do not drain pods that are using local node storage in emptyDir",
 	},
-	"dry-run": map[string]interface{}{
-		"key":      dryRunConfigKey,
-		"defValue": false,
-		"usage":    "If true, only log if a node would be drained",
+	"dry-run": flagData{
+		Key:      dryRunConfigKey,
+		DefValue: false,
+		Usage:    "If true, only log if a node would be drained",
 	},
-	"enable-scheduled-event-draining": map[string]interface{}{
-		"key":      enableScheduledEventDrainingConfigKey,
-		"defValue": enableScheduledEventDrainingDefault,
-		"usage":    "[EXPERIMENTAL] If true, drain nodes before the maintenance window starts for an EC2 instance scheduled event",
+	"enable-scheduled-event-draining": flagData{
+		Key:      enableScheduledEventDrainingConfigKey,
+		DefValue: enableScheduledEventDrainingDefault,
+		Usage:    "[EXPERIMENTAL] If true, drain nodes before the maintenance window starts for an EC2 instance scheduled event",
 	},
-	"enable-spot-interruption-draining": map[string]interface{}{
-		"key":      enableSpotInterruptionDrainingConfigKey,
-		"defValue": enableSpotInterruptionDrainingDefault,
-		"usage":    "If true, drain nodes when the spot interruption termination notice is receieved",
+	"enable-spot-interruption-draining": flagData{
+		Key:      enableSpotInterruptionDrainingConfigKey,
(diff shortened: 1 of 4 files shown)


pkg/config/config.go
```diff
@@ -23,70 +25,151 @@ import (
...
-	flag.StringVar(&config.WebhookTemplate, "webhook-template", getEnv(webhookTemplateConfigKey, webhookTemplateDefault), "If specified, replaces the default webhook message template.")
-	flag.BoolVar(&config.EnableScheduledEventDraining, "enable-scheduled-event-draining", getBoolEnv(enableScheduledEventDrainingConfigKey, enableScheduledEventDrainingDefault), "[EXPERIMENTAL] If true, drain nodes before the maintenance window starts for an EC2 instance scheduled event")
-	flag.BoolVar(&config.EnableSpotInterruptionDraining, "enable-spot-interruption-draining", getBoolEnv(enableSpotInterruptionDrainingConfigKey, enableSpotInterruptionDrainingDefault), "If true, drain nodes when the spot interruption termination notice is receieved")
 
-	flag.Parse()
+	results, err := createFlags(flagData)
+	if err != nil {
+		return config, err
+	}
+	gracePeriod := results["grace-period"].(int)
```
bwagner5: this would panic if grace-period is not an int. You'll probably want to wrap this in a .(type) switch with int as a case and default being the error case.

pkg/config/config.go
```diff
@@ -23,70 +25,151 @@ import (
...
+	NodeName                       string `json:"node-name"`
+	NodeTerminationGracePeriod     int    `json:"node-termination-grace-period"`
+	PodTerminationGracePeriod      int    `json:"pod-termination-grace-period"`
+	WebhookURL                     string `json:"webhook-url"`
+	WebhookHeaders                 string `json:"webhook-headers"`
+	WebhookTemplate                string `json:"webhook-template"`
+}
+
+var flagData = map[string]map[string]interface{}{
+	"delete-local-data": map[string]interface{}{
```
bwagner5: instead  of defining an interface{} for each of these, could we pull that out to a struct with key, defValue, and usage as fields? defValue would still be of interface{} type but key and usage could be typed to string and then you wouldn't have to hardcode the fields when you're looking things up below.

pkg/config/config.go
```diff
@@ -130,7 +213,41 @@ func ParseCliArgs() Config {
...
+	return config, nil
+}
+
+func createFlags(flagData map[string]map[string]interface{}) (map[string]interface{}, error) {
+	result := make(map[string]interface{})
+
+	for name, data := range flagData {
+		switch data["defValue"].(type) {
+		case string:
+			value := getEnv(data["key"].(string), data["defValue"].(string))
```
bwagner5: using a typed key field in a struct would also prevent a possible panic here if "key" for some reason was not a string.

pkg/config/config_internal_test.go
```diff
@@ -0,0 +1,185 @@
...
+	"testing"
+
+	h "github.com/aws/aws-node-termination-handler/pkg/test"
+)
+
+// All of these needed for TestIsConfigProvided
+var location string
+var cliArgName = "name"
+var envVarName = "NAME_TEST"
+var value = "haugenj"
```
bwagner5: nice 😁

pkg/config/config.go
```diff
@@ -144,29 +261,29 @@ func getEnv(key string, fallback string) string {
...
+func getIntEnv(key string, fallback int) (int, error) {
 	envStrValue := getEnv(key, "")
 	if envStrValue == "" {
-		return fallback
+		return fallback, nil
 	}
 	envIntValue, err := strconv.Atoi(envStrValue)
 	if err != nil {
-		log.Fatalln("Env Var " + key + " must be an integer")
+		return -1, err
```
bwagner5: instead of blindly passing back the err can you use `fmt.Errorf("Env var %s must be an integer: %w", key, err)` to wrap the error message to give some context where the Atoi conversion failed.

bwagner5: or errors.New is fine too, like below

pkg/config/config.go
jillmon: It will and I can, but createFlags will ensure that whatever the user inputs is the same type as the default value for grace-period. The default value for grace-period is -1, so we'll never get this far if the user input for grace-period isn't an int. I can still add the checks though. It'll probably make people feel better.

bwagner5: ah i c, hadn't gotten that far yet. Makes sense :)

Issue #74, opened 2020-01-28
cannot use when restricting access to metadata service


cdaniluk:
In order to prevent pods from assuming the IAM role of the node, we restrict access to the metadata service for pods as follows:

```
yum install -y iptables-services
iptables --insert FORWARD 1 --in-interface eni+ --destination 169.254.169.254/32 --jump DROP
iptables-save | tee /etc/sysconfig/iptables 
systemctl enable --now iptables
```

This is recommended in the [AWS EKS documentation](https://docs.aws.amazon.com/eks/latest/userguide/restrict-ec2-credential-access.html) for worker nodes.

When running the termination handler, the following is logged:

```
2020/01/28 00:33:47 Request to instance metadata failed. Retrying.
2020/01/28 00:34:17 Error getting response from instance metadata  Get http://169.254.169.254/latest/meta-data/spot/instance-action: dial tcp 169.254.169.254:80: i/o timeout
```

Mostly logging this so that other people who run into the same know this isn't supported.. but are there any plans to add support for gathering the termination events through CloudWatch events when restricting access to the metadata service? Or should we just use an alternate termination handler?

mattrandallbecker:
Definitely makes sense if you block metadata this handler would not work.

Looks like that documentation recommends that you block the entirety of metadata service via a userdata script such that it's not accessible on the host.  You could not use an instance profile and not assign the metadata credentials any permissions. Then you wouldn't need to prevent assuming the role or block metadata.

Alternatively, we are investigating support for the [instance metadata store v2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) which has authentication, you could then use a secret store or configure the handler to use the token to get metadata information. https://github.com/aws/aws-node-termination-handler/issues/22

Lastly, we did investigate supporting CloudWatch events as a means of getting instance notifications but that would require a different approach than this repo currently employs.  Likely via a centralized controller listening for events rather than a distributed daemonset, but we wanted to avoid having to set up any AWS specific permissions to get the project to work for ease of use.

marcincuber:
I have the same scenario as @cdaniluk. I think this is a massive issue with this aws-node-termination-handler.

mattrandallbecker:
@marcincuber 

[...]

PR #85, opened 2020-01-31
adding hostnetwork to daemonset pod spec

mattrandallbecker:
Issue #, if available:
https://github.com/aws/aws-node-termination-handler/issues/74

Description of changes:
if following the EKS documentation [here](https://docs.aws.amazon.com/eks/latest/userguide/restrict-ec2-credential-access.html) customers would not be able to use this handler as it prevents forwarding metadata, changing the pod spec to hostnetwork allows the daemonset pod to continue to get information from metadata service


mattrandallbecker:
ok so it appears host level networking cant call services, this is a problem

mattrandallbecker:
also we still need this change to support IMDSv2 so i'll have to change the integration tests to make this work properly

Issue #22, opened 2019-11-21
Support IMDSv2


bwagner5:
Support the recently released instance metadata service v2 (https://aws.amazon.com/about-aws/whats-new/2019/11/announcing-updates-amazon-ec2-instance-metadata-service/)

bwagner5:
Here's some more info on IMDSv2:

https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/

Re:Invent 2019: 
https://www.youtube.com/watch?v=2B5bhZzayjI

bwagner5:
the code to support IMDSv2 has been merged, but I'll wait to close this issue until the release has been pushed.

bwagner5:
released in v1.3.0

PR #87, opened 2020-02-03
imdsv2 support

bwagner5:
Issue #, if available:
https://github.com/aws/aws-node-termination-handler/issues/22

Description of changes:
Support IMDSv2 w/ an IMDSv1 fallback. There is no configuration since imdsv2 should always be favored, but will fallback to imdsv1 if there is an issue or if it's blocked due to a container hop if not configured with a TTL of 2. 

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -24,6 +24,7 @@ import (
 	"github.com/aws/aws-node-termination-handler/pkg/config"
 	"github.com/aws/aws-node-termination-handler/pkg/drainevent"
 	"github.com/aws/aws-node-termination-handler/pkg/draineventstore"
+	"github.com/aws/aws-node-termination-handler/pkg/ec2metadata"
 	"github.com/aws/aws-node-termination-handler/pkg/node"
 	"github.com/aws/aws-node-termination-handler/pkg/webhook"
 )
@@ -31,9 +32,10 @@ import (
 const (
 	scheduledMaintenance = "Scheduled Maintenance"
 	spotITN              = "Spot ITN"
+	metadataTries        = 3
 )
 
-type monitorFunc func(chan<- drainevent.DrainEvent, chan<- drainevent.DrainEvent, config.Config) error
+type monitorFunc func(chan<- drainevent.DrainEvent, chan<- drainevent.DrainEvent, *ec2metadata.EC2MetadataService) error
 
 func main() {
 	signalChan := make(chan os.Signal, 1)
@@ -68,6 +70,8 @@ func main() {
 	cancelChan := make(chan drainevent.DrainEvent)
 	defer close(cancelChan)
 
+	imds := ec2metadata.New(nthConfig.MetadataURL, metadataTries)
+
 	monitoringFns := map[string]monitorFunc{}
 	if nthConfig.EnableSpotInterruptionDraining {
 		monitoringFns[spotITN] = drainevent.MonitorForSpotITNEvents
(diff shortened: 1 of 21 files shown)


cmd/node-termination-handler.go
```diff
@@ -24,16 +24,18 @@ import (
...
 	"github.com/aws/aws-node-termination-handler/pkg/draineventstore"
+	"github.com/aws/aws-node-termination-handler/pkg/ec2metadata"
 	"github.com/aws/aws-node-termination-handler/pkg/node"
 	"github.com/aws/aws-node-termination-handler/pkg/webhook"
 )
 
 const (
 	scheduledMaintenance = "Scheduled Maintenance"
 	spotITN              = "Spot ITN"
+	metadataTries        = 3
```
mattrandallbecker: how do you feel about moving this outside this file? feels like its part of configuration unrelated to which monitoring functions we are tracking

config/helm/ec2-metadata-test-proxy/values.yaml
```diff
@@ -7,6 +7,7 @@ ec2MetadataTestProxy:
     scheduledEventStatus: active
     enableScheduledMaintenanceEvents: false
     enableSpotITN: true
+    enableIMDSV2: false
```
mattrandallbecker: why is this a value at all? it doesn't seem used in the code and even if it's set to false we seemingly try v2 in each request unless im missing something

pkg/drainevent/scheduled-event.go
```diff
@@ -56,17 +54,11 @@ func MonitorForScheduledEvents(drainChan chan<- DrainEvent, cancelChan chan<- Dr
 }
 
 // checkForScheduledEvents Checks EC2 instance metadata for a scheduled event requiring a node drain
-func checkForScheduledEvents(metadataURL string) ([]DrainEvent, error) {
-	resp, err := ec2metadata.RequestMetadata(metadataURL, ec2metadata.ScheduledEventPath)
+func checkForScheduledEvents(imds *ec2metadata.EC2MetadataService) ([]DrainEvent, error) {
+	scheduledEvents, err := imds.GetScheduledMaintenanceEvents()
```
mattrandallbecker: this is so much better

pkg/ec2metadata/ec2metadata.go
```diff
@@ -67,12 +89,135 @@ type InstanceAction struct {
...
+		tries:       tries,
+		httpClient:  http.Client{},
+	}
+}
+
+// GetScheduledMaintenanceEvents retrieves EC2 scheduled maintenance events from imds
+func (e *EC2MetadataService) GetScheduledMaintenanceEvents() ([]ScheduledEventDetail, error) {
+	resp, err := e.Request(ScheduledEventPath)
+	if resp != nil && (resp.StatusCode < 200 || resp.StatusCode >= 300) {
+		return nil, fmt.Errorf("Unable to request metadata, received http status code: %d", resp.StatusCode)
```
mattrandallbecker: nit, we were able to request metadata, we just didnt like the response

test/k8s-local-cluster-test/run-test
```diff
@@ -185,8 +193,11 @@ export EC2_METADATA_DOCKER_REPO=$(echo $EC2_METADATA_DOCKER_IMG | cut -d':' -f1)
 export EC2_METADATA_DOCKER_TAG=$(echo $EC2_METADATA_DOCKER_IMG | cut -d':' -f2)
 ###
 
+i=0
 for assert_script in $ASSERTION_SCRIPTS; do
   reset_cluster
+  export IMDS_PORT=$(expr $i + 1338)
```
mattrandallbecker: much smarter port collision mechanism, thank you

pkg/drainevent/scheduled-event.go
bwagner5: yep, that's a good idea

cmd/node-termination-handler.go
bwagner5: yep, that's a good idea

PR #91, opened 2020-02-04
removing security context from chart and updating tag version

mattrandallbecker:
Issue #, if available:
N/A

Description of changes:
podSecurityContext was in readme and values, but not in any chart, so i removed it
securityContext - privileged was set to true when it didnt need to be
values image.tag 1.1.0 -> 1.2.0


jillmon:
https://github.com/aws/eks-charts/pull/66

PR #92, opened 2020-02-05
Add standalone yaml templates and update docs

marcincuber:
Description of changes:

Re-add standalone YAML templates that can used to deploy aws-node-termination-handler without using HELM. 

Personally, I am using Flux GitOps operator which deploys and syncs yaml templates for me and therefore I am not using helm at all.

### DIFF
--- a/config/deploy/base/daemonset.yaml
+++ b/config/deploy/base/daemonset.yaml
@@ -1,0 +1,84 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    app: aws-node-termination-handler
+  name: aws-node-termination-handler
+  namespace: kube-system
+spec:
+  selector:
+    matchLabels:
+      app: aws-node-termination-handler
+  template:
+    metadata:
+      labels:
+        app: aws-node-termination-handler
+    spec:
+      nodeSelector:
+        lifecycle: Ec2Spot
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: beta.kubernetes.io/os
+                operator: In
+                values:
+                - linux
+              - key: beta.kubernetes.io/arch
+                operator: In
+                values:
+                - amd64
+      containers:
+      - env:
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              apiVersion: v1
+              fieldPath: spec.nodeName
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              apiVersion: v1
+              fieldPath: metadata.name
+        - name: NAMESPACE
+          valueFrom:
+            fieldRef:
+              apiVersion: v1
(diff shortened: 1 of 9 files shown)


config/deploy/base/daemonset.yaml
```diff
@@ -0,0 +1,84 @@
...
+              fieldPath: status.podIP
+        - name: DELETE_LOCAL_DATA
+          value: "true"
+        - name: IGNORE_DAEMON_SETS
+          value: "true"
+        - name: NODE_TERMINATION_GRACE_PERIOD
+          value: "120"
+        - name: POD_TERMINATION_GRACE_PERIOD
+          value: "10"
+        image: amazon/aws-node-termination-handler:v1.2.0
```
jaypipes: How do you propose ensuring that this value stays in sync with the Helm value?

https://github.com/aws/aws-node-termination-handler/blob/9fa5b501a4b99450109755943992aeb0e9f86135/config/helm/aws-node-termination-handler/values.yaml#L7

config/deploy/base/daemonset.yaml
```diff
@@ -0,0 +1,84 @@
...
+        imagePullPolicy: IfNotPresent
+        name: aws-node-termination-handler
+        resources:
+          limits:
+            cpu: 100m
+            memory: 128Mi
+          requests:
+            cpu: 50m
+            memory: 64Mi
```
jaypipes: or that these values don't get out of sync with the Helm values?

https://github.com/aws/aws-node-termination-handler/blob/9fa5b501a4b99450109755943992aeb0e9f86135/config/helm/aws-node-termination-handler/values.yaml#L17-L23

marcincuber:
@bwagner5 I think adding a generation process in Travis would be a good solution and it would definitely work for me. 
Kustomize file is not a massive thing for me but it is built into kubectl so it is worth making use of it. Most importantly I would like to see simple yaml templates that can be used with `kubectl apply` and just work.

bwagner5:
opened an issue to track the progress of providing plain yaml resource files rather than requiring helm use: https://github.com/aws/aws-node-termination-handler/issues/93

mattrandallbecker:
sorry for leaving this long enough that it got stale, we discussed and are going to try and make the template generation part of the build and release process so that yaml files are included automatically with every release

bwagner5:
@marcincuber let me know if this PR https://github.com/aws/aws-node-termination-handler/pull/96 takes care of your use case.  I've backfilled the kubernetes yaml resource files in the github release artifacts for each version. There is an aggregate yaml file that includes all resources in one yaml file and then a tar-ball that includes the individual yaml resource files that you could easily drop a kustomize file into. 

https://github.com/aws/aws-node-termination-handler/releases

marcincuber:
@bwagner5 That is very nice and definitely fits my bill. Just a small note, it would be good to remove helm specific annotations etc. 
Otherwise, we can close this PR.

mattrandallbecker:
I've opened the following issues for us to fix that in the future https://github.com/aws/aws-node-termination-handler/issues/100

Issue #93, opened 2020-02-06
Provide k8s yaml resource files in addition to helm chart


bwagner5:
Some users would rather use the Kubernetes resource yaml files directly rather than installing via helm charts. This presents a problem with synchronizing the resource files and the helm chart. 

We already store a local copy of the helm chart in this repo and the distributed copy in the github.com/aws/eks-charts repo. The synchronization between these two locations is done manually but enforced via a travis test which diffs the two locations. 

A possible solution is to treat the helm chart as the source of truth and use `helm template` to generate the yaml resource files and then upload as a tar'd artifact in our github releases. This could all be done via travis (we're already automatically generating binaries and uploading to releases via travis). The generation process would ensure that the default values are the same between the plain-yaml files and the helm default values.

PR #96, opened 2020-02-14
build and upload k8s resource files from helm as the base

bwagner5:
Issue #, if available:
https://github.com/aws/aws-node-termination-handler/issues/93

Description of changes:
- Script the generation of plain kubernetes resources yaml files to upload as assets to the github release. This change supports users who would rather not use helm to install aws-node-termination-handler. 

- I also set the default in helm for psp generation to true. afaik, generating the psp won't harm anything even if rbac is not enabled in the cluster, but correct me if I'm wrong. 


Makefile
```diff
@@ -59,10 +59,13 @@ build-binaries:
...
 	${MAKEFILE_PATH}/scripts/upload-binaries-to-github
 
+build-and-upload-k8s-resources:
+	${MAKEFILE_PATH}/scripts/build-and-upload-k8s-resources-yaml
+
 sync-readme-to-dockerhub:
 	${MAKEFILE_PATH}/scripts/sync-readme-to-dockerhub
 
-release: create-build-dir build-binaries upload-binaries-to-github
+release: create-build-dir build-binaries upload-binaries-to-github build-and-upload-k8s-resources
```
mattrandallbecker: as discussed offline can we make the release process | create -> build binaries -> build yaml -> upload

scripts/generate-k8s-yaml
```diff
@@ -0,0 +1,33 @@
...
+AGG_RESOURCES_YAML=$BUILD_DIR/aws-node-termination-handler-$VERSION.yaml
+mkdir -p $BUILD_DIR/indv-resources
+
+curl -L https://get.helm.sh/helm-v$HELM_VERSION-$PLATFORM-amd64.tar.gz | tar zxf - -C $BUILD_DIR
+mv $BUILD_DIR/$PLATFORM-amd64/helm $BUILD_DIR/.
+rm -rf $BUILD_DIR/$PLATFORM-amd64
+chmod +x $BUILD_DIR/helm
+
+$BUILD_DIR/helm template aws-node-termination-handler \
+    --namespace kube-system \
```
jaypipes: nit: might be worth parameterizing the namespace instead of hard-coding it to kube-system?

README.md
```diff
@@ -67,6 +67,11 @@ helm upgrade --install aws-node-termination-handler \
   eks/aws-node-termination-handler
 ```
 
+Basic installation without Helm:
+```sh
+kubectl apply -f https://github.com/aws/aws-node-termination-handler/releases/download/v1.2.0/aws-node-termination-handler-v1.2.0.yaml
```
jaypipes: small suggestion... I realize why you did it this way (because of how you're generating the files), but having the aws-node-termination-handler name and version in the filename in addition to the releases download URL seems a bit verbose.

In other words, instead of this:

```
kubectl apply -f https://github.com/aws/aws-node-termination-handler/releases/download/v1.2.0/aws-node-termination-handler-v1.2.0.yaml 
```

I think it would be nicer to do this:

```
kubectl apply -f https://github.com/aws/aws-node-termination-handler/releases/download/v1.2.0/all.yaml

Comment (bwagner5) [reply README.md]: agreed, I moved the version identifier from the filenames to a directory in the build dir. This should prevent confusion between generation and upload scripts while keeping the asset URLs clean.

File: scripts/upload-resources-to-github [hunk 4]
```diff
@@ -3,6 +3,11 @@ set -euo pipefail
 
 SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
 VERSION=$(make -s -f $SCRIPTPATH/../Makefile version)
+BUILD_DIR=$SCRIPTPATH/../build/k8s-resources/$VERSION
+INDV_RESOURCES_DIR=$BUILD_DIR/indv-resources
+TAR_RESOURCES_FILE=$BUILD_DIR/aws-node-termination-handler-k8s-resources.tar
+AGG_RESOURCES_YAML=$BUILD_DIR/aws-node-termination-handler.yaml
```
jaypipes: I still think it would be a little more cogent to name the tarball `individual-resources.tar` and the aggregated YAML file `all.yaml` or `all-resources.yaml`. Thoughts?

bwagner5: ah yes, I see what you mean. I overlooked the additional stutter in your first comment. I'll update

PR #116, opened 2020-03-10
restrict filesystem and volume writes from container and run as non-root

bwagner5:
Issue #, if available:
N/A

Description of changes:
Locks down some access from NTH:
 - Makes the root container filesystem read only
 - Runs as non-root
 - Disallow privilege escalation explicitly
 - change uptime volume mount to readonly


bwagner5:
There's a pending PR in eks-charts to sync the repos, so I'm waiting for that to go through before opening up a PR to sync this change. The only test failing is the eks-charts in-sync test.

PR #122, opened 2020-03-17
changing text proxy to avoid integ test webhook spam

mattrandallbecker:
Issue #, if available:
N/A

Description of changes:
spot itn time is now fixed in the metadata proxy

### DIFF
--- a/test/ec2-metadata-test-proxy/cmd/ec2-metadata-test-proxy.go
+++ b/test/ec2-metadata-test-proxy/cmd/ec2-metadata-test-proxy.go
@@ -51,7 +51,7 @@ const (
 )
 
 var startTime int64 = time.Now().Unix()
-var spotInterruptionTime string = time.Now().Local().Add(time.Minute * time.Duration(2)).Format(time.RFC3339)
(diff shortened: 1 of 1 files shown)


test/ec2-metadata-test-proxy/cmd/ec2-metadata-test-proxy.go
```diff
@@ -51,6 +51,7 @@ const (
 )
 
 var startTime int64 = time.Now().Unix()
+var spotInterruptionTime string = time.Now().Local().Add(time.Minute * time.Duration(2)).Format(time.RFC3339)
```
bwagner5: should use `time.Now().UTC()...` rather than `Local()`

bwagner5 (changes requested):

test/ec2-metadata-test-proxy/cmd/ec2-metadata-test-proxy.go
bwagner5: I realize now that I wrote that original line which you just moved up.... how embarrassing. But it should indeed be UTC() rather than Local()

mattrandallbecker:
i changed the other two places that file used Local() as well for maintenance events

bwagner5 (approved):
LGTM

Issue #138, opened 2020-04-15
RBAC question: DaemonSet delete


coderanger:
Your sample deployment manifest includes delete permissions for daemonsets. Is this actually used? It looks like maybe it's only needed for running your test suite, and not for normal usage?

mattrandallbecker:
You may be right that these permissions are overly broad. We'll do some testing to see if we can remove them.

coderanger:
I just went looking for any actual calls to a Delete API in the code and didn't find any, but they might be hiding in a library or similar :)

bwagner5:
We've merged the unneeded permissions removal in this repo, but waiting on https://github.com/aws/eks-charts/pull/114 PR to be merged and an eks-charts release.

bwagner5:
released https://github.com/aws/aws-node-termination-handler/releases/tag/v1.4.0

bwagner5:
released https://github.com/aws/aws-node-termination-handler/releases/tag/v1.4.0

PR #140, opened 2020-04-16
Remove unnecessary delete daemonset permission

haugenj:
Issue #138, if available:

Description of changes:

Deleted the the line in the cluster role config file that gives NTH permission to delete daemonsets. 

### DIFF
--- a/config/helm/aws-node-termination-handler/templates/clusterrole.yaml
+++ b/config/helm/aws-node-termination-handler/templates/clusterrole.yaml
@@ -36,4 +36,3 @@ rules:
     - daemonsets
   verbs:
     - get
-    - delete
--- a/config/helm/aws-node-termination-handler/templates/clusterrole.yaml
+++ b/config/helm/aws-node-termination-handler/templates/clusterrole.yaml
@@ -26,7 +26,6 @@ rules:
 - apiGroups:
     - extensions
   resources:
-    - replicasets
     - daemonsets
   verbs:
     - get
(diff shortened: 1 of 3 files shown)


config/helm/aws-node-termination-handler/templates/clusterrole.yaml
```diff
@@ -36,4 +36,3 @@ rules:
     - daemonsets
   verbs:
     - get
```
jaypipes, line 38: Actually, why does the node termination handler need *any* access to daemonset or replicaset resources? AFAIK, NTH only [queries Node](https://github.com/aws/aws-node-termination-handler/blob/4fe54278740ca7174d50d8f4c7892da55c2219ad/pkg/node/node.go#L313)?

haugenj: Good point, I'll test it out removing those as well and see if there's any gotchas

haugenj: Just a small update, removing the get daemonset permissions is causing the e2e test to break so I'm looking into what's going on there. I'm not totally convinced it's a real error rather than a problem with the test

haugenj: It looks like the daemonset get permissions are required so that we can filter out the daemonsets within the RunNodeDrain call: https://github.com/kubernetes/kubernetes/blob/fb6f5d739b08e049b188806d6e33099f85fad2fe/staging/src/k8s.io/kubectl/pkg/drain/drain.go#L189

I haven't had any issues with removing the replicaset permissions though, so I'll take that out in a new revision.

haugenj:
Corresponding change to the eks-chart https://github.com/aws/eks-charts/pull/114

PR #155, opened 2020-05-07
Revert date format in human-readable logs

haugenj:
Issue #, if available: none

Description of changes:

This reverts the date format in the logs back to how it was before PR https://github.com/aws/aws-node-termination-handler/pull/152. This doesn't affect the ability to use JSON logging, it only changes the human-readable format. 

The thinking is that it was wrong to modify the format in a minor release, so we'll revisit changing the log format when we're close to version 2.0.0.

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -33,13 +33,14 @@ import (
 const (
 	scheduledMaintenance = "Scheduled Maintenance"
 	spotITN              = "Spot ITN"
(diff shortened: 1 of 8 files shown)


bwagner5:
looks like unit-tests and go-report-card failed for the same problem:

```
could not parse "vet: pkg/webhook/webhook_test.go:49:64:error: too many arguments (vet)" - strconv.Atoi: parsing " pkg/webhook/webhook_test.go": invalid syntax
```

PR #158, opened 2020-05-11
set idle conns timeout on webhook http client

bwagner5:
Issue #, if available:
N/A

Description of changes:
The newly added webhook-http-proxy-test was failing when testing with an actual https endpoint because the squid access log was not being written until the connection was closed. However, the go http client was keeping the connection open in the connection pool for reuse. There's no reason to reuse the http connection since the webhook is a one-time request and most often the instance is about to be terminated. 

- Set IdleConnTimeout to 1 second on the webhook http client
- Switch Squid from daemon logging to stdio logging for instant flush
- Use the full docker image hash for squid 


jaypipes:
"Switch Squid from daemon logging to stdio logging for instant flush" <-- it's best practice generally to just log to stdout for containers anyway, and handle log shipping through a separate exporter container or host utility.

PR #166, opened 2020-05-21
update travis to run license test on tag releases

bwagner5:
Issue #, if available:
N/A

Description of changes:
Travis should run license test on tag releases too. 


brycahta:
Changes look good, but not merging because of failed checks in Travis. Can you take a look?

bwagner5:
hmmm this must be intermittent failures due to the taint. I noticed that coredns is being evicted since the master is also being tainted. I think I just need to prevent the ec2-metadata-test-proxy from being scheduled on the master node.

Issue #147, opened 2020-04-30
Unable to parsed scheduled event start time


yzargari:
Hello,
I recently deployed aws-node-termination-handler in my cluster and now I have my first scheduled termination. The node hasn't drain nor cordoned and the application logs show the following message repeatedly every 2-3 seconds:
```
2020/04/30 12:19:25 There was a problem monitoring for Scheduled Maintenance events: Unable to parsed scheduled event start time: parsing time "8 May 2020 16:00:00 GMT" as "02 Jan 2006 15:04:05 GMT": cannot parse "8 May 2020 16:00:00 GMT" as "02"
2020/04/30 12:19:27 There was a problem monitoring for Scheduled Maintenance events: Unable to parsed scheduled event start time: parsing time "8 May 2020 16:00:00 GMT" as "02 Jan 2006 15:04:05 GMT": cannot parse "8 May 2020 16:00:00 GMT" as "02"
2020/04/30 12:19:29 There was a problem monitoring for Scheduled Maintenance events: Unable to parsed scheduled event start time: parsing time "8 May 2020 16:00:00 GMT" as "02 Jan 2006 15:04:05 GMT": cannot parse "8 May 2020 16:00:00 GMT" as "02"
2020/04/30 12:19:32 There was a problem monitoring for Scheduled Maintenance events: Unable to parsed scheduled event start time: parsing time "8 May 2020 16:00:00 GMT" as "02 Jan 2006 15:04:05 GMT": cannot parse "8 May 2020 16:00:00 GMT" as "02"
2020/04/30 12:19:34 There was a problem monitoring for Scheduled Maintenance events: Unable to parsed scheduled event start time: parsing time "8 May 2020 16:00:00 GMT" as "02 Jan 2006 15:04:05 GMT": cannot parse "8 May 2020 16:00:00 GMT" as "02"
2020/04/30 12:19:35 There was a problem monitoring for Scheduled Maintenance events: Unable to parsed scheduled event start time: parsing time "8 May 2020 16:00:00 GMT" as "02 Jan 2006 15:04:05 GMT": cannot parse "8 May 2020 16:00:00 GMT" as "02"
2020/04/30 12:19:37 There was a problem monitoring for Scheduled Maintenance events: Unable to parsed scheduled event start time: parsing time "8 May 2020 16:00:00 GMT" as "02 Jan 2006 15:04:05 GMT": cannot parse "8 May 2020 16:00:00 GMT" as "02"
[...]

PR #182, opened 2020-06-25
do not fail when scheduled event does not have end date

bwagner5:
Overview:
When processing scheduled maintenance events, do not fail when scheduled event does not have an end date. This information is not present for all events and is not utilized in the functionality of aws-node-termination-handler. It only serves as additional, nice-to-have information for logging/webhook msg purposes. 

Issue #, if available:
https://github.com/aws/aws-node-termination-handler/issues/147

Description of changes:
 - Made end-time parsing optional for scheduled event processing. Still logging that the parsing failed, but it will now continue to send the event over the channel for potential cordon+drain.


haugenj:
👍

Issue #8, opened 2019-11-08
Support windows spot instances


rparsonsbb:
Looks like the current daemonset and image is only for linux, is there a plan on the roadmap to support windows spot instances or should a pull request be made?

mattrandallbecker:
We'll definitely investigate windows support.  We'll do some testing and get back to you.

doryanmzr:
any update on that?

mattrandallbecker:
Sorry for the delay, we got closer to supporting windows with our manifest builds for docker images but we still need to prioritize the build and release script changes to make this happen then perform some testing to ensure it works.

doryanmzr:
> Sorry for the delay, we got closer to supporting windows with our manifest builds for docker images but we still need to prioritize the build and release script changes to make this happen then perform some testing to ensure it works.

I'd like to know if there is any progress :)

cjerad:
> I'd like to know if there is any progress

I'm working on this now

cshivashankar:
Hi, Any idea when this might be released. I am trying to get an approximate time frame so that I can plan accordingly.

jillmon:
@cjerad is currently in the process of scoping the requirements for this change. We should be able to give you a better estimate of completion within another week or so. Since this is such a new feature, we want to make sure that it is adequately reviewed and tested, but this feature is now on our official roadmap. We expect to complete it soon.

cshivashankar:
Hi @jillmon Many thanks for the update. I will be following this thread for more info

bwagner5:
Windows support was released a while back, thanks to @cjerad!  🥳🎊🚀🤩

We would love to hear feedback from anyone using NTH on windows!  I'm going to close this issue, but feel free to comment here on your experience or open a new issue.

PR #185, opened 2020-06-29
add Windows node support

cjerad:
Issue #, if available:
Resolves aws/aws-node-termination-handler#8

Description of changes:
Adds support for running on Windows nodes.  Also updates build and test scripts.

### DIFF
--- a/scripts/run-unit-tests-in-docker
+++ b/scripts/run-unit-tests-in-docker
@@ -1,0 +1,71 @@
+#!/bin/bash
+
+set -euo pipefail
+
+project_root_dir="$(cd "$(dirname "$0")/.." && pwd -P)"
+work_dir="/workplace/aws-node-termination-handler"
+container_name="nth_unit_test_on_linux"
+deps="go,git,make"
+recreate=0
+usage=$(cat <<EOM
+  usage: run-unit-tests-in-docker [-c CONTAINER_NAME] [-d DEP1,...] [-r]
+
+    Runs "make unit-test" in an Alpine Linux docker container.
+
+    Creates and initializes a new container if one does not already exist.
+
+    Options:
+      -c     Container name (default "$container_name")
+      -d     Linux dependencies to install (default "$deps")
+      -r     Recreate container
+
+EOM
+)
+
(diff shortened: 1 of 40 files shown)


scripts/run-unit-tests-in-docker
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+set -euo pipefail
+
+project_root_dir="$(cd "$(dirname "$0")/.." && pwd -P)"
+work_dir="/workplace/aws-node-termination-handler"
+container_name="nth_unit_test_on_linux"
+deps="go,git,make"
```
bwagner5, line 8: instead of using alpine and installing dependencies, you can use the golang:1 docker container which already has go, git, and make installed.

scripts/run-unit-tests-in-docker
```diff
@@ -0,0 +1,71 @@
...
+  docker container rm "$container_name" >/dev/null 2>&1 || true
+fi
+
+if ! [[ -n $(docker container ls -a | grep "$container_name") ]]; then
+  echo "creating container ..."
+
+  IFS=',' read -ra deps <<< "$deps"
+  echo "dependencies to install: ${deps[@]}"
+
+  docker container create \
```
bwagner5, line 54: I think the default workflow would be to run the unit tests and discard the container.  Maybe change "recreate" to "preserve" instead which would effectively remove the `--rm` from the command below. 

`docker run -it --rm -e GOPROXY=direct --volume "$project_root_dir:$work_dir" --workdir "$work_dir" --name "$container_name" golang:1`

yurrriq:
I'm getting the following error since upgrading from 0.8.0 to 0.9.2. Seems potentially related to this PR. I have not set that value.

```
Error: Failed to render chart: exit status 1: Error: template: aws-node-termination-handler/templates/daemonset.linux.yaml:28:85: executing "aws-node-termination-handler/templates/daemonset.linux.yaml" at <.Values.linuxPodAnnotations>: wrong type for value; expected map[string]interface {}; got interface {}
```

PR #192, opened 2020-07-08
Misc changes to aws-node-termination-handler Helm chart

yurrriq:
Description of changes:
A few typos and the like I noticed. Let me know if you'd like me to create separate pull requests.

It's also unclear to me whether the PR should go here or to https://github.com/aws/eks-charts/

### DIFF
--- a/config/helm/aws-node-termination-handler/Chart.yaml
+++ b/config/helm/aws-node-termination-handler/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 name: aws-node-termination-handler
 description: A Helm chart for the AWS Node Termination Handler
 version: 0.9.1
-appVersion: 1.6.0
+appVersion: 1.6.1
 home: https://github.com/aws/eks-charts
 icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
 sources:
--- a/config/helm/aws-node-termination-handler/README.md
+++ b/config/helm/aws-node-termination-handler/README.md
@@ -57,7 +57,7 @@ Parameter | Description | Default
 `gracePeriod` | (DEPRECATED: Renamed to podTerminationGracePeriod) The time in seconds given to each pod to terminate gracefully. If negative, the default value specified in the pod will be used. | `30`
 `podTerminationGracePeriod` | The time in seconds given to each pod to terminate gracefully. If negative, the default value specified in the pod will be used. | `30`
 `nodeTerminationGracePeriod` | Period of time in seconds given to each NODE to terminate gracefully. Node draining will be scheduled based on this value to optimize the amount of compute time, but still safely drain the node before an event. | `120`
-`ignoreDaemonsSets` | Causes kubectl to skip daemon set managed pods | `true`
+`ignoreDaemonSets` | Causes kubectl to skip daemon set managed pods | `true`
 `instanceMetadataURL` | The URL of EC2 instance metadata. This shouldn't need to be changed unless you are testing. | `http://169.254.169.254:80`
 `webhookURL` | Posts event data to URL upon instance interruption action | ``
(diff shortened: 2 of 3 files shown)


yurrriq:
It seems misleading and perhaps a bit of a maintenance issue to have default values `""` and then mention what I presume to be program defaults in the chart README. What's the thinking there? Perhaps in such cases we should at least omit the variables from the DaemonSet manifest.

config/helm/aws-node-termination-handler/values.yaml
```diff
@@ -30,14 +30,13 @@ resources:
     memory: "64Mi"
     cpu: "50m"
   limits:
-    memory: "128Mi"
-    cpu: "100m"
```
yurrriq, line 34: e.g. https://medium.com/omio-engineering/cpu-limits-and-aggressive-throttling-in-kubernetes-c5b20bd8a718

bwagner5: This was specifically related to the Linux CFS scheduler bug which has been patched and back ported to 4.18.  aws-node-termination-handler doesn't require much in terms of cpu or memory anyways, so I doubt the bug would have impacted NTH. The limits could probably match the request tbh. If the CFS bug did or still can impact NTH, a pause wouldn't be that bad for NTH, since it's just responding to events and is required to act in seconds, not milliseconds.  The limits provide a nice safety mechanism in the case of a bug introduced in the codebase, so I don't think the cons of having it outweigh that benefit.

config/helm/aws-node-termination-handler/README.md
```diff
@@ -57,7 +57,7 @@ Parameter | Description | Default
 `gracePeriod` | (DEPRECATED: Renamed to podTerminationGracePeriod) The time in seconds given to each pod to terminate gracefully. If negative, the default value specified in the pod will be used. | `30`
 `podTerminationGracePeriod` | The time in seconds given to each pod to terminate gracefully. If negative, the default value specified in the pod will be used. | `30`
 `nodeTerminationGracePeriod` | Period of time in seconds given to each NODE to terminate gracefully. Node draining will be scheduled based on this value to optimize the amount of compute time, but still safely drain the node before an event. | `120`
-`ignoreDaemonsSets` | Causes kubectl to skip daemon set managed pods | `true`
+`ignoreDaemonSets` | Causes kubectl to skip daemon set managed pods | `true`
```
bwagner5, line 60: nice catch! Can you submit this in another PR?

config/helm/aws-node-termination-handler/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 name: aws-node-termination-handler
 description: A Helm chart for the AWS Node Termination Handler
 version: 0.9.1
-appVersion: 1.6.0
+appVersion: 1.6.1
```
bwagner5, line 5: the app version should stay 1.6.0, the chart version should be incremented when only modifying the helm chart.

yurrriq: Heh yep. I must've absentmindedly incremented the wrong one.

config/helm/aws-node-termination-handler/README.md
yurrriq: https://github.com/aws/aws-node-termination-handler/pull/199

yurrriq:
> Having said that, I think we can improve here through better automation. Maybe a fix that makes this easier for users to understand and easier for the maintainers to keep up with is a test that checks if the CLI config defaults match the helm default values files and the helm readme? Would that be something you'd be interested in contributing? Maybe we could open an issue to track?
> 
> I'm hesitant to merge this because it's so easy to copy over the wrong defaults. For example, the `enableSpotInterruptionDraining` is defaulted to `true` in the CLI, but it was copied over in this PR as false. In this case, the end-to-end tests caught the error, but there could be another case that isn't explicitly checked by e2e tests that may fall through the cracks.

I fully agree. I close this and maybe take a stab at better automation. Thanks for the CFS comments too. Makes sense.

PR #198, opened 2020-07-10
add shellcheck and spellcheck

bwagner5:
*Issue #, if available:*
N/A

*Description of changes:*
 - Add Shellcheck to check for errors in bash files
    - Fix errors found in bash files
 - Add spellcheck to check markdown files
    - Fix spelling errors in markdown files

### DIFF
--- a/test/k8s-local-cluster-test/provision-cluster
+++ b/test/k8s-local-cluster-test/provision-cluster
@@ -85,7 +85,7 @@ for dep in "${deps[@]}"; do
     if [ ! -x "$path_to_executable" ]; then
         echoerr "You are required to have $dep installed on your system..."
         echoerr "Please install $dep and try again. "
-        exit -1
+        exit 3
     fi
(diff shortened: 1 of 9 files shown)


test/k8s-local-cluster-test/provision-cluster
```diff
@@ -85,7 +85,7 @@ for dep in "${deps[@]}"; do
     if [ ! -x "$path_to_executable" ]; then
         echoerr "You are required to have $dep installed on your system..."
         echoerr "Please install $dep and try again. "
-        exit -1
+        exit 3
```
haugenj, line 88: y tho

bwagner5: was using bash exit codes improperly, apparently exit only accepts positive integers and exit 0, 1, 2 and supposed to be special-ish, so starting custom error codes at 3 :)

PR #201, opened 2020-07-14
refactor event handlers to monitors with an interface

bwagner5:
Issue #, if available:
N/A

Description of changes:
- This PR refactors event handler code to use a Monitor interface instead of being called by function name. This will enable more complicated monitors to be created in the future ;) 

- This PR also updates the node pkg to accept the nodeName as input to functions that need to lookup or manipulate a node. 


pkg/interruptioneventstore/interruption-event-store_test.go
```diff
@@ -26,13 +26,19 @@ import (
 	h "github.com/aws/aws-node-termination-handler/pkg/test"
 )
 
+const (
+	node1 = "test-node-1"
+	node2 = "test-node-2"
```
haugenj, line 31: unused?

pkg/interruptionevent/spot-itn-monitor_test.go
```diff
@@ -60,7 +60,9 @@ func TestMonitorForSpotITNEventsSuccess(t *testing.T) {
 			"Expected description to contain: "+startTime+" but is actually: "+result.Description)
 	}()
 
-	err := interruptionevent.MonitorForSpotITNEvents(drainChan, cancelChan, imds)
+	nodeName := "test-node"
```
haugenj, line 63: can we make this a constant?

cmd/node-termination-handler.go
```diff
@@ -87,25 +87,27 @@ func main() {
...
-		monitoringFns[scheduledMaintenance] = interruptionevent.MonitorForScheduledEvents
+		imdsScheduledEventMonitor := interruptionevent.NewScheduledEventMonitor(imds, interruptionChan, cancelChan, nthConfig.NodeName)
+		monitoringFns[scheduledMaintenance] = imdsScheduledEventMonitor
 	}
 
-	for eventType, fn := range monitoringFns {
-		go func(monitorFn monitorFunc, eventType string) {
-			log.Log().Msgf("Started monitoring for %s events", eventType)
+	for _, fn := range monitoringFns {
+		go func(monitor interruptionevent.Monitor) {
```
haugenj, line 101: is `type monitorFunc ...` on line 40 necessary anymore?

pkg/interruptionevent/scheduled-event-monitor_test.go
```diff
@@ -61,6 +61,7 @@ func TestMonitorForScheduledEventsSuccess(t *testing.T) {
 	drainChan := make(chan interruptionevent.InterruptionEvent)
 	cancelChan := make(chan interruptionevent.InterruptionEvent)
 	imds := ec2metadata.New(server.URL, 1)
+	nodeName := "test-node"
```
haugenj, line 64: const?

pkg/interruptionevent/spot-itn-monitor_test.go
```diff
@@ -60,7 +60,9 @@ func TestMonitorForSpotITNEventsSuccess(t *testing.T) {
...
+	spotITNMonitor := interruptionevent.NewSpotInterruptionMonitor(imds, drainChan, cancelChan, nodeName)
+	err := spotITNMonitor.Monitor()
```
haugenj, line 65: this is pedantic but should we change these function names? They're referencing functions that are removed. `TestMonitor` would be better imo

pkg/interruptionevent/spot-itn-monitor_test.go
bwagner5: yees

pkg/interruptioneventstore/interruption-event-store_test.go
bwagner5: unused

cmd/node-termination-handler.go
bwagner5: nope

pkg/interruptionevent/scheduled-event-monitor_test.go
bwagner5: yees

pkg/interruptionevent/spot-itn-monitor_test.go
bwagner5: the problem is that they're still in the same package, so those names would conflict

bwagner5: I was thinking about moving these handler into individual sub packages... idk if that's too much though, but it would isolate them.  Ideally, they shouldn't share anything. wdut?

haugenj: That sounds good to me

bwagner5: updated

Issue #202, opened 2020-07-15
Can't use nodeSelector with helm chart


kuzaxak:
```
helm upgrade --install aws-node-termination-handler \
  --namespace kube-system \
  --set nodeSelector.lifecycle=spot \
  eks/aws-node-termination-handler
```

Gives you an error: 
```
Error: template: aws-node-termination-handler/templates/daemonset.linux.yaml:157:26: executing "aws-node-termination-handler/templates/daemonset.linux.yaml" at <8>: wrong type for value; expected string; got map[string]interface {}
```

Version:
```
helm3 version
version.BuildInfo{Version:"v3.2.4", GitCommit:"0ad800ef43d3b826f31a5ad8dfbb4fe05d143688", GitTreeState:"clean", GoVersion:"go1.13.12"}
```

haugenj:
Thanks for bringing this up!

We released chart version 1.6 last night that includes support for Windows, and after a quick look it seems we might not be passing the value for nodeSelector correctly. We'll investigate further as soon as we can

bwagner5:
released fix in eks-charts

PR #203, opened 2020-07-15
fix helm chart nodeSelector (#202)

cjerad:
Issue #202

**Description**
Correctly convert value into YAML.

Verified that generating chart template with `--set nodeSelector.lifecycle=spot,nodeSelector.foo=bar` outputs:

```yaml
nodeSelector:
    kubernetes.io/os: linux
    foo: bar
    lifecycle: spot
```

### DIFF
--- a/config/helm/aws-node-termination-handler/templates/daemonset.linux.yaml
+++ b/config/helm/aws-node-termination-handler/templates/daemonset.linux.yaml
@@ -154,10 +154,10 @@ spec:
       nodeSelector:
         {{ include "aws-node-termination-handler.nodeSelectorTermsOs" . }}: linux
         {{- with .Values.nodeSelector }}
-          {{- . | nindent 8 }}
+          {{- toYaml . | nindent 8 }}
         {{- end }}
         {{- with .Values.linuxNodeSelector }}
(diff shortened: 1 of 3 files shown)


config/helm/aws-node-termination-handler/templates/daemonset.linux.yaml
```diff
@@ -154,10 +154,10 @@ spec:
       nodeSelector:
         {{ include "aws-node-termination-handler.nodeSelectorTermsOs" . }}: linux
         {{- with .Values.nodeSelector }}
-          {{- . | nindent 8 }}
+          {{- toYaml . | nindent 8 }}
```
jillmon, line 157: What was it converting the value to previously?

cjerad: It just directly inserted the string.

denniswebb:
@bwagner5 When will this be pushed to the public repository? Latest there is `0.9.1`?

bwagner5:
@denniswebb PR is out on the eks-charts repo https://github.com/aws/eks-charts/pull/212  should be released soon

PR #221, opened 2020-08-04
default enableSpotInterruptionDraining to true

toneill818:
Description of changes:
According to the README, the default value for enableSpotInterruptionDraining is true, but the value in values.yaml had it set to empty string (""). 


haugenj:
Thanks for submitting this! Without a value set for enableSpotInterruptionDraining in the Helm chart we use the default value in [config.go](https://github.com/aws/aws-node-termination-handler/blob/master/pkg/config/config.go#L51) which is set to true.

I think rather than modify the default value in the Helm chart off of "" which would be inconsistent with the other Helm values, it would be better for us to improve the wording of the comment so it's more clear that true is indeed the default value.

PR #242, opened 2020-08-18
clear up shellcheck warnings

bwagner5:
Issue #, if available:
N/A

Description of changes:

There have been a lot of build/test/deployment script changes lately and have caused some problems since these scripts are difficult to test locally. As an enhanced safety measure, which would've caught some of these errors, this PR clears up shellcheck warnings and now only passes if shell scripts do not contain warnings. Before, shellcheck was only set to fail on errors. 

### DIFF
--- a/scripts/push-docker-images
+++ b/scripts/push-docker-images
@@ -67,7 +67,7 @@ if [[ $MANIFEST == "true" ]]; then
     echo "manifest already exists"
     EXISTING_IMAGES=()
     while IFS='' read -r line; do 
-      shell_files+=("$line"); 
+      EXISTING_IMAGES+=("$line"); 
     done < <(docker manifest inspect $IMAGE_REPO:$VERSION | jq -r '.manifests[] | "\(.platform.os)-\(.platform.architecture)"')
     # treat separate from PLATFORMS because existing images don't need to be tagged and pushed
     for os_arch in "${EXISTING_IMAGES[@]}"; do
--- a/scripts/push-docker-images
+++ b/scripts/push-docker-images
@@ -60,18 +60,21 @@ if [[ $MANIFEST == "true" ]]; then
     echo '{"experimental":"enabled"}' > $DOCKER_CLI_CONFIG
     echo "Created docker config file"
   fi
-  cat <<< $(jq '.+{"experimental":"enabled"}' $DOCKER_CLI_CONFIG) > $DOCKER_CLI_CONFIG
+  cat <<< "$(jq '.+{"experimental":"enabled"}' $DOCKER_CLI_CONFIG)" > $DOCKER_CLI_CONFIG
   echo "Enabled experimental CLI features to execute docker manifest commands"
   manifest_exists=$(docker manifest inspect $IMAGE_REPO:$VERSION > /dev/null ; echo $?)
   if [[ manifest_exists -eq 0 ]]; then
     echo "manifest already exists"
-    EXISTING_IMAGES=($(docker manifest inspect $IMAGE_REPO:$VERSION | jq -r '.manifests[] | "\(.platform.os)-\(.platform.architecture)"'))
+    EXISTING_IMAGES=()
+    while IFS='' read -r line; do 
+      shell_files+=("$line"); 
+    done < <(docker manifest inspect $IMAGE_REPO:$VERSION | jq -r '.manifests[] | "\(.platform.os)-\(.platform.architecture)"')
     # treat separate from PLATFORMS because existing images don't need to be tagged and pushed
     for os_arch in "${EXISTING_IMAGES[@]}"; do
       img_tag="$IMAGE_REPO:$VERSION-$os_arch"
-      MANIFEST_IMAGES+=($img_tag)
+      MANIFEST_IMAGES+=("$img_tag")
     done
-    echo "images already in manifest: $MANIFEST_IMAGES"
+    echo "images already in manifest: ${MANIFEST_IMAGES[*]}"
   fi
 fi
 
@@ -88,7 +91,7 @@ for os_arch in "${PLATFORMS[@]}"; do
         docker tag $img_tag_w_platform $img_tag
     fi
     docker push $img_tag
-    MANIFEST_IMAGES+=($img_tag)
+    MANIFEST_IMAGES+=("$img_tag")
 done
 
 if [[ $MANIFEST == "true" ]]; then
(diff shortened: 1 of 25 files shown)


scripts/push-docker-images
```diff
@@ -60,18 +60,21 @@ if [[ $MANIFEST == "true" ]]; then
...
-  cat <<< $(jq '.+{"experimental":"enabled"}' $DOCKER_CLI_CONFIG) > $DOCKER_CLI_CONFIG
+  cat <<< "$(jq '.+{"experimental":"enabled"}' $DOCKER_CLI_CONFIG)" > $DOCKER_CLI_CONFIG
   echo "Enabled experimental CLI features to execute docker manifest commands"
   manifest_exists=$(docker manifest inspect $IMAGE_REPO:$VERSION > /dev/null ; echo $?)
   if [[ manifest_exists -eq 0 ]]; then
     echo "manifest already exists"
-    EXISTING_IMAGES=($(docker manifest inspect $IMAGE_REPO:$VERSION | jq -r '.manifests[] | "\(.platform.os)-\(.platform.architecture)"'))
+    EXISTING_IMAGES=()
+    while IFS='' read -r line; do 
+      shell_files+=("$line");
```
brycahta, line 70: Can you explain this a bit? Not sure why it needs to be here

test/e2e/imds-v2-test
```diff
@@ -94,6 +94,7 @@ for i in `seq 1 10`; do
         DEPLOYED=1
         break
     fi
+    echo "Setup Loop $i/10, sleeping for 5 seconds"
```
brycahta, line 97: Could use `$SETUP_SLEEP` to be consistent with `$TAINT_CHECK_SLEEP`

test/e2e/maintenance-event-dry-run-test
```diff
@@ -101,6 +101,7 @@ for i in $(seq 0 10); do
       echo "✅ Scheduled Maintenance Event Dry Run Test Passed $CLUSTER_NAME! ✅"
       exit 0
   fi
+    echo "Assertion Loop $i/10, sleeping for 10 seconds"
```
brycahta, line 104: do you think it's worth going with the larger sleep value to cover all tests and be consistent?

scripts/push-docker-images
bwagner5: it was recommended here: https://github.com/koalaman/shellcheck/wiki/SC2207

It basically ensures that the array additions are added as separate elements without splitting the splitting globs mid-line.

test/e2e/prometheus-metrics-test
```diff
@@ -139,9 +140,9 @@ fi
 POD_NAME=$(get_nth_worker_pod)
 echo "✅ Fetched the pod $POD_NAME "
 
-kubectl -n kube-system port-forward $POD_NAME 7000:9092 &
+kubectl -n kube-system port-forward "$POD_NAME" 7000:9092 &
 PORT_FORWARD_PID=$!
-trap "kill ${PORT_FORWARD_PID}" EXIT SIGINT SIGTERM ERR
+trap 'kill ${PORT_FORWARD_PID}' EXIT SIGINT SIGTERM ERR
```
brycahta, line 145: just curious, do you remember why it recommends single quote here?

scripts/push-docker-images
brycahta: oh or copy/paste error on the var? prob should be `EXISTING_IMAGES` right?

test/helm-sync-test/run-helm-sync-test
```diff
@@ -12,8 +13,16 @@ mkdir -p $TMP_DIR
...
+
+branches=()
+while IFS='' read -r line; do 
+   branches+=("$line"); 
+done < <(echo $prs | jq -r '.[].head.ref')
+
+clone_urls=()
+while IFS='' read -r line; do 
+   clone_urls+=("$line"); 
+done < <(echo $prs | jq -r '.[].head.repo.clone_url')
```
brycahta, line 25: that sloc thoo 👀  👀

test/shellcheck/run-shellcheck
```diff
@@ -14,10 +14,14 @@ function exit_and_fail() {
...
+curl -Lo ${BUILD_DIR}/shellcheck.tar.xz "https://github.com/koalaman/shellcheck/releases/download/v${SHELLCHECK_VERSION}/shellcheck-v${SHELLCHECK_VERSION}.${KERNEL}.x86_64.tar.xz"
 tar -C ${BUILD_DIR} -xvf "${BUILD_DIR}/shellcheck.tar.xz"
 export PATH="${BUILD_DIR}/shellcheck-v${SHELLCHECK_VERSION}:$PATH"
 
-shellcheck -S error $(grep -Rn -e '#!.*/bin/bash' -e '#!.*/usr/bin/env bash' ${SCRIPTPATH}/../../ | grep ":[1-5]:" | cut -d':' -f1)
+shell_files=()
+while IFS='' read -r line; do 
+   shell_files+=("$line"); 
+done < <(grep -Rnl -e '#!.*/bin/bash' -e '#!.*/usr/bin/env bash' ${SCRIPTPATH}/../../)
+shellcheck -S warning "${shell_files[@]}"
```
brycahta, line 25: high res photo of the code quality bar:  📈

love to see it

test/e2e/prometheus-metrics-test
bwagner5: it's to make sure the expansion happens on the trap signal rather than at definition. In this case, it doesn't matter, so I went with the recommendation rather than the shellcheck disable line above.

test/e2e/maintenance-event-dry-run-test
bwagner5: yeah, this didn't even make any sense. I was being lazy, yet doing so, led to more work and now rework 🤷‍♂️

scripts/push-docker-images
bwagner5: GOOD CATCH

test/e2e/imds-v2-test
bwagner5: yeah, there's really no reason it can't be the same as the TAINT_CHECK_SLEEP for ultimate consistency

test/e2e/cordon-only-test
```diff
@@ -89,6 +89,7 @@ for i in `seq 1 10`; do
         DEPLOYED=1
         break
     fi
+    echo "Setup Loop $i/10, sleeping for 5 seconds"
```
brycahta, line 92: missed a spot

Issue #229, opened 2020-08-11
Unable to parse metadata response: Metadata request received http status code: 401


benzch:
Hello, 

I recently had retirement event to one of my worker nodes. I noticed this worker isn't draining, I found this error in the logs 

`2020/08/07 07:32:28 There was a problem monitoring for Scheduled Maintenance events: Unable to parse metadata response: Metadata request received http status code: 401`

This error occur every second 
Result from worker :

`curl 169.254.169.254/latest/meta-data/events/maintenance/scheduled
[ {
  "NotBefore" : "21 Aug 2020 10:00:00 GMT",
  "Code" : "instance-stop",
  "Description" : "The instance is running on degraded hardware",
  "EventId" : "instance-event-07c59a40db2a78c8a",
  "State" : "active"
} ]`

tcpdump from host on one of the failed queries: 

13:38:04.793607 IP (tos 0x0, ttl 255, id 6105, offset 0, flags [DF], proto TCP (6), length 60)
    10.51.144.143.61069 > 169.254.169.254.http: Flags [S], cksum 0xeeed (incorrect -> 0xbc86), seq 4040972178, win 26883, options [mss 8961,sackOK,TS val 1824485981 ecr 0,nop,wscale 7], length 0
E..<..@...u#
3.........P..W.......i.......#....
l.r]........
13:38:04.793672 IP (tos 0x0, ttl 255, id 0, offset 0, flags [DF], proto TCP (6), length 60)
    169.254.169.254.http > 10.51.144.143.61069: Flags [S.], cksum 0x5176 (correct), seq 1744321961, ack 4040972179, win 27084, options [mss 9040,sackOK,TS val 868454530 ecr 1824485981,nop,wscale 7], length 0
E..<..@.........
3...P..g.=...W...i.Qv....#P...
3...l.r]....
13:38:04.793695 IP (tos 0x0, ttl 255, id 6106, offset 0, flags [DF], proto TCP (6), length 52)
    10.51.144.143.61069 > 169.254.169.254.http: Flags [.], cksum 0xeee5 (incorrect -> 0x06d8), seq 4040972179, ack 1744321962, win 211, options [nop,nop,TS val 1824485981 ecr 868454530], length 0
E..4..@...u*
3.........P..W.g.=............
l.r]3...
13:38:04.793789 IP (tos 0x0, ttl 255, id 6107, offset 0, flags [DF], proto TCP (6), length 277)
[...]

PR #244, opened 2020-08-20
Add retries when reponse from IMDSv2 retruns a 401

DP19:
Issue #, if available:
#229 
Description of changes:
Added a for loop for getting scheduled maintenance events and spot instance events. Set the retry limit to 1. 

Not sure of the best way to add this to the test suite. Also curious if throwing a panic to sovle the second item in the issue is what's best to do here or not. 

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -114,7 +114,7 @@ func main() {
 						duplicateErrCount = 0
 						previousErr = err
 					}
-					if duplicateErrCount > duplicateErrThreshold {
+					if duplicateErrCount >= duplicateErrThreshold {
 						log.Log().Msg("Stopping NITH - Duplicate Error Threshold hit.")
 						panic(fmt.Sprintf("%v",err))
 					}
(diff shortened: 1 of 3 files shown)


cmd/node-termination-handler.go
```diff
@@ -100,11 +101,23 @@ func main() {
...
 					log.Log().Msgf("There was a problem monitoring for %s events: %v", monitor.Kind(), err)
 					metrics.ErrorEventsInc(monitor.Kind())
+					if err == previousErr {
+						duplicateErrCount++
+					} else {
+						duplicateErrCount = 0
+						previousErr = err
+					}
+					if duplicateErrCount >= duplicateErrThreshold {
+						log.Log().Msg("Stopping NITH - Duplicate Error Threshold hit.")
```
bwagner5, line 118: Can you remove the "I" in NTH. 

"Stopping NTH - ..."

DP19: Updated

PR #248, opened 2020-08-27
upgrade test dependency versions

bwagner5:
Issue #, if available:
N/A

Description of changes:
 - Upgrade helm to latest versions
 - Add k8s 1.19 kind docker image to e2e tests
 - patch version upgrade of k8s 1.18 kind docker image
 - Change default k8s version to 1.17
 - Remove k8s 1.12 and 1.13 since Kind no longer supports those version (latest supported eks is 1.14 so seems fine)
 - Enable rbac on AEMM

### DIFF
--- a/test/e2e/cordon-only-test
+++ b/test/e2e/cordon-only-test
@@ -52,7 +52,6 @@ aemm_helm_args=(
   "$AEMM_DL_URL"
   --namespace default
   --set servicePort="$IMDS_PORT"
-  --set 'rbac.pspEnabled=true'
 )
 [[ ${#common_helm_args[@]} -gt 0 ]] &&
     aemm_helm_args+=("${common_helm_args[@]}")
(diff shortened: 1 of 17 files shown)


bwagner5:
Will wait for kind to officially support k8s 1.19 before merging

test/e2e/cordon-only-test
```diff
@@ -52,6 +52,7 @@ aemm_helm_args=(
   "$AEMM_DL_URL"
   --namespace default
   --set servicePort="$IMDS_PORT"
+  --set 'rbac.pspEnabled=true'
```
haugenj, line 55: why do we need to set this to true in all the tests now?

bwagner5: not needed, I just needed to update the cluster psp since k8s 1.19 seems to default to seccomp to `runtime/default` rather than `docker/default`

haugenj (approved):
lgtm

PR #253, opened 2020-09-14
Allow users to configure webhook message content with a template file

supasteev0:
Description of changes:

When the destination of the webhook requires a more complex message content, it can be easier to provide a json template file which will be mounted into the pod, rather than having the webhook message content set in an inline variable.

This will allow users to set a WEBHOOK_TEMPLATE_FILE environment variable (--webhook-template-file argument) to set the template file path which will be used to send the webhook message. This takes precedence over the --webhook-template argument.

Template file will have to exist (be mounted into the pod as volume by users) otherwise it will throw an error.

Example: 
```
---
apiVersion: v1
data:
  webhook-alert-template.json: |-
    {
      "status": "firing",
      "alerts": [
          {
            "status": "firing",
            "labels": {
              "alertname": "myalert",
              "region": "eu-west-1",
              "severity": "high"
            }
          }
       ]
    }
kind: ConfigMap
metadata:
  name: webhook-template
  namespace: monitoring
---
apiVersion: apps/v1
kind: DaemonSet
metadata:
  labels:
    app: aws-node-termination-handler
  namespace: monitoring
spec:
  selector:
    matchLabels:
      app: aws-node-termination-handler
  template:
    metadata:
      labels:
        app: aws-node-termination-handler
    spec:
      containers:
      - env:
        ...
        - name: WEBHOOK_TEMPLATE_FILE
          value: /config/webhook-alert-template.json
        ...
        volumeMounts:
        - mountPath: /config/
          name: template
      volumes:
      - configMap:
          name: webhook-template
        name: template
```

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -17,6 +17,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"text/template"
@@ -35,8 +36,21 @@ type combinedDrainData struct {
 
 // Post makes a http post to send drain event data to webhook url
 func Post(additionalInfo ec2metadata.NodeMetadata, event *monitor.InterruptionEvent, nthconfig config.Config) {
+	var webhookTemplateContent = ""
 
-	webhookTemplate, err := template.New("message").Parse(nthconfig.WebhookTemplate)
+	if nthconfig.WebhookTemplateFile != "" {
+		content,err := ioutil.ReadFile(nthconfig.WebhookTemplateFile)
+		if err != nil {
+			log.Log().Msgf("Webhook Error: Template parsing failed - %s", err)
+			return
+		}
+		webhookTemplateContent = string(content)
+		log.Log().Msgf("Template file content - %s", webhookTemplateContent)
+	} else {
+		webhookTemplateContent = nthconfig.WebhookTemplate
+	}
+
+	webhookTemplate, err := template.New("message").Parse(webhookTemplateContent)
 	if err != nil {
 		log.Log().Msgf("Webhook Error: Template parsing failed - %s", err)
 		return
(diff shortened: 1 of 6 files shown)


bwagner5 (changes requested):
Rather than making the user mount the file, can you instead use a ConfigMap similar to the way the `webhookURLSecretName` works.

supasteev0:
@bwagner5 I have update the Helm chart by adding 2 new variables in order to set the configmap name and the template file name that will be mounted into the pod.

pkg/webhook/webhook.go
```diff
@@ -35,8 +36,21 @@ type combinedDrainData struct {
...
+	var webhookTemplateContent string
+
+	if nthconfig.WebhookTemplateFile != "" {
+		content, err := ioutil.ReadFile(nthconfig.WebhookTemplateFile)
+		if err != nil {
+			log.Log().Msgf("Webhook Error: Could not read template file %s - %s", nthconfig.WebhookTemplateFile, err)
+			return
+		}
+		webhookTemplateContent = string(content)
+		log.Log().Msgf("Template file content - %s", webhookTemplateContent)
```
bwagner5, line 48: can you move this to`log.Debug() .Msgf(`

bwagner5 (changes requested):
This looks good! I missed one thing during my first pass review.

Can you update the `ValidateWebhookConfig()` function as well? The validate function is called on startup so you know if your template is invalid upfront rather than waiting for a termination event for the error to be thrown. 

https://github.com/aws/aws-node-termination-handler/blob/master/pkg/webhook/webhook.go#L103

pkg/webhook/webhook.go
supasteev0: Done

supasteev0:
I updated the ValidateWebhookConfig function as requested. I also fixed a typo on the nthConfig variable which was written in 2 different ways. Looks like the travis test has failed because of an issue with ec2Mock not responding, but I don't know how to re trigger the job.

bwagner5 (approved):
LGTM! 🚀 thanks for the contribution!

Issue #249, opened 2020-09-01
JSON logging isn't truly JSON logging


kemra102:
We have JSON logging enabled. But the messages are basically plain text (not quite truly Human readable). For example here is the log for a spot termination:

```
Got interruption event from channel {InstanceID:i-0cb9fad8f8e327c4c InstanceType:m5d.xlarge PublicHostname:ec2-54-76-23-97.eu-west-1.compute.amazonaws.com PublicIP:54.76.23.97 LocalHostname:ip-10-200-205-41.aeuw1.env.product.cloud.zone LocalIP:10.200.205.41} {EventID:spot-itn-5ef7cf326c13be727fb52feb7ff7ee5658b36478d4697ab81603a2e727a7bb99 Kind:SPOT_ITN Description:Spot ITN received. Instance will be interrupted at 2020-09-01T09:44:02Z 
 State: StartTime:2020-09-01 09:44:02 +0000 UTC EndTime:0001-01-01 00:00:00 +0000 UTC Drained:false PreDrainTask:0x10cc2f0}
```

This is a string, with a regular string upfront and 2 embedded JSON data structures on the end. As such, automated JSON log parsing in products such as Logstash/ElasticSearch don't work. So even after turning on JSON logging if we wanted to properly parse these log entries we'd still need a custom transformation rule in Logstash/ES to do so. Which kind of defeats the point in my opinion.

This is just one example, other parts of the application don't appear to even try to do JSON logging, such as startup config printing for example.

I think making the entire app "truly" support JSON logging is a huge effort and I'm not sure how to go about it or even if this is desirable to the project maintainers.

haugenj:
Hey @kemra102 !

I agree the JSON logging could be better. When we added JSON logging with Zerolog we wanted to make sure the default human-readable logs displayed in exactly the same format as they were before, in case anyone was scraping the logs and parsing them already. However, like you brought up this limits the JSON log format and keeps it from really being JSON logging. Since that time we've added Prometheus metrics to NTH so changing the format might not be as much of an issue anymore. 

I haven't worked with Logstash or ElasticSearch, so if we did change the format here what would be the ideal output? Anything different than the following?
```
{
"message" : "Got interruption event from channel",
"time" : "2020-09-03T00:00:00Z",
"level" : "info",
"metadata" :
  {
  "InstanceID" : "i-12345678",
  "InstanceType" : "t3.large",
  ...
  },
"event" :
  {
[...]

PR #255, opened 2020-09-14
Reformat logs to support proper JSON logging

haugenj:
Issue #, if available:
#249 

Description of changes:
Reformat logs so that they actually use json formatting for the errors and fields instead of hiding those values within the 'message' field.

**Question** - should we make the log key format more consistent? 
- Standard keys like 'message', 'time', and 'error' are lowercase due to zerolog defaults
- The arguments are logged in kebab case to match the old format. 
- Custom keys like 'NodeName' are pascal case because I matched them against how .Interface() values are logged (they match their struct names which are pascal case) 

I'm learning toward making all the keys we control snake_case.

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -85,7 +85,7 @@ func NewWithValues(nthConfig config.Config, drainHelper *drain.Helper, uptime up
 // CordonAndDrain will cordon the node and evict pods based on the config
 func (n Node) CordonAndDrain(nodeName string) error {
 	if n.nthConfig.DryRun {
-		log.Log().Str("NodeName", nodeName).Msg("Node would have been cordoned and drained, but dry-run flag was set")
+		log.Log().Str("node_name", nodeName).Msg("Node would have been cordoned and drained, but dry-run flag was set")
 		return nil
 	}
 	log.Log().Msg("Cordoning the node")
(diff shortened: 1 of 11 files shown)


pkg/node/node.go
```diff
@@ -471,18 +473,28 @@ func addTaint(node *corev1.Node, nth Node, taintKey string, taintValue string, e
 		}
 
 		if err != nil {
-			log.Log().Msgf("Error while adding %v taint on node %v: %v", taintKey, node.Name, err)
+			log.Log().
```
brycahta, line 476: out of curiosity, what does this log look like compared to the one you have posted? I'm assuming you didn't wrap in `Err` for formatting reasons

log.Log().Err("Error while adding taint on node").Str("TaintKey", taintKey).  ...

pkg/node/node.go
```diff
@@ -535,10 +549,16 @@ func removeTaint(node *corev1.Node, client kubernetes.Interface, taintKey string
 		}
 
 		if err != nil {
-			log.Log().Msgf("Error while releasing %v taint on node %v: %v", taintKey, node.Name, err)
+			log.Log().
+				Str("TaintKey", taintKey).
+				Str("NodeName", node.Name).
+				Msg("Error while releasing taint on node")
```
brycahta, line 555: I like this format in code, makes it easier to read/understand 👍🏼

brycahta:
> I'm learning toward making all the keys we control snake_case.

**+ 1**

brycahta (approved):
LGTM 🎆

pkg/node/node.go
haugenj: what do you mean by "compared to the one you have posted"? 

I thought wrapping our message in `Err()` would be misleading since the actual error is stored in `err`. So in the logs it would look like there are two errors when there's really only one. I removed `err` from the log statement here for a similar reason - since we're returning `err` it should be handled by the calling function and logging it in both places seemed redundant.

brycahta: Got it, I just meant comparing what you currently have in code now vs. the same line but wrapping the message with `Err()`

PR #257, opened 2020-09-21
Retry uncordon workflow to handle transient network issues

nithu0115:
Issue #, if available: N/A

Description of changes: 

Recently, we had an edge case scenario where NTH daemon failed to uncordon the node with below error message:

```
"Unable to complete the uncordon after reboot workflow on startup: Unable to fetch kubernetes node from API: Get " https://172.20.0.1:443/api/v1/nodes/ip-10-79-29-243.us-east-2.compute.internal": dial tcp 172.20.0.1:443: i/o timeout errors. 
```

Debugging further, we could neither find any issue with kube-proxy d/s or aws-node d/s which constantly talks to API server upon which we thought this could be a transient network issue due to which NTH failed to uncordon the node. 

That said, from a distributed systems stand point, we should be able to handle any such network partition events for at least short period of time without evicting the whole cluster and node termination handler should not worry about such issues as Kubelet will send heart beats to api server which will automatically mark the node unschedulable if that is the case. 

This change introduces retries on such transient network issues. It retries 4 times in 8 seconds period at an interval of 2 seconds. 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+.idea/
 
 # Output of the go build for the cmd binary
 /node-termination-handler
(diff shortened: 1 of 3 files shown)


bwagner5 (approved):
LGTM! Thanks! 🚀

PR #259, opened 2020-09-25
Log all pods on a node when cordoning the node

wanwenli:
### Issue #, if available

N.A.

### Description of changes:

To log all pods that are potentially affected by a spot interruption.

### Decisions

1. Logging the pod names can be another `PreDrainTask`, but that would require the PreDrainTask of `monitor.InterruptionEvent` to be changed to a slice and much more code changes. On the other hand, for a scheduled maintenance, there is no need to log all pod names.
1. Make `LogPods ` and `fetchAllPods` a member function of Node such that they can potentially be reused.
1. Change a bit in the test cases to see if the logging works properly.

✅  By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/aws/aws-node-termination-handler/pkg/config"
 	"github.com/aws/aws-node-termination-handler/pkg/uptime"
+	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -286,6 +287,21 @@ func (n Node) TaintSpotItn(nodeName string, eventID string) error {
 	return addTaint(k8sNode, n, SpotInterruptionTaint, eventID, corev1.TaintEffectNoSchedule)
 }
 
+// LogPods logs all the pod names on a node
+func (n Node) LogPods(nodeName string) error {
+	podList, err := n.fetchAllPods(nodeName)
+	if err != nil {
+		return fmt.Errorf("Unable to fetch all pods from API: %w", err)
+	}
+	podNamesArr := zerolog.Arr()
+	for _, pod := range podList.Items {
+		podNamesArr = podNamesArr.Str(pod.Name)
+	}
+	log.Log().Array("pod_names", podNamesArr).Str("node_name", nodeName).Msg("Pods on node")
+
+	return nil
+}
+
 // TaintScheduledMaintenance adds the scheduled maintenance taint onto a node
 func (n Node) TaintScheduledMaintenance(nodeName string, eventID string) error {
 	if !n.nthConfig.TaintNode {
@@ -397,6 +413,12 @@ func (n Node) fetchKubernetesNode(nodeName string) (*corev1.Node, error) {
 		return node, nil
 	}
 	return n.drainHelper.Client.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
+}
+
+func (n Node) fetchAllPods(nodeName string) (*corev1.PodList, error) {
+	return n.drainHelper.Client.CoreV1().Pods("").List(metav1.ListOptions{
+		FieldSelector: "spec.nodeName=" + nodeName,
+	})
 }
 
 func getDrainHelper(nthConfig config.Config) (*drain.Helper, error) {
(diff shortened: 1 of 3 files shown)


haugenj:
Hey @wanwenli, thanks for the PR! 

Can you elaborate on the motivation for this change? Pods are already logged when they're evicted (or ignored) as part of the Drain API and I'm curious what the limitations of this are, and why we shouldn't implement this for scheduled maintenance events as well. 
```
2020/09/25 19:59:47 Got interruption event from channel {InstanceID: InstanceType: PublicHostname: PublicIP: LocalHostname:ip-123-456-789-123.ec2.internal LocalIP:123.456.789.123 AvailabilityZone:us-east-1a} {EventID:spot-itn-0846f7580d8093ddfc87b20fe57f69639bca9e9fb2163de9eb30ca31faf7e9fe Kind:SPOT_ITN Description:Spot ITN received. Instance will be interrupted at 2020-09-25T20:01:47Z
 State: NodeName:ip-123-456-789-1231.us-east-2.compute.internal StartTime:2020-09-25 20:01:47 +0000 UTC EndTime:0001-01-01 00:00:00 +0000 UTC Drained:false PreDrainTask:0x113c8a0 PostDrainTask:<nil>}
WARNING: ignoring DaemonSet-managed Pods: default/amazon-ec2-metadata-mock-bqdq9, kube-system/aws-node-bl2bj, kube-system/aws-node-termination-handler-dzjfk, kube-system/kube-proxy-fct9f
evicting pod "coredns-67bfd975c5-gvzjs"
evicting pod "kube-ops-view-78d7db44dd-vmz7v"
evicting pod "coredns-67bfd975c5-6hwfj"
```
(the format of these improves with the release of #255)

wanwenli:
Hey @haugenj, your point is very valid. I have changed the place where pod names are logged to when cordon takes place. In this way, scheduled maintenance makes the pod names to be logged in the case of cordoning a node.

pkg/node/node.go
```diff
@@ -286,6 +287,21 @@ func (n Node) TaintSpotItn(nodeName string, eventID string) error {
...
+func (n Node) LogPods(nodeName string) error {
+	podList, err := n.fetchAllPods(nodeName)
+	if err != nil {
+		return fmt.Errorf("Unable to fetch all pods from API: %w", err)
+	}
+	podNamesArr := zerolog.Arr()
+	for _, pod := range podList.Items {
+		podNamesArr = podNamesArr.Str(pod.Name)
+	}
+	log.Log().Array("pod_names", podNamesArr).Msgf("Pods on node %s", nodeName)
```
haugenj, line 300: Can you log the node name with `Str("node_name", nodeName)` instead?

pkg/node/node.go
```diff
@@ -286,6 +287,21 @@ func (n Node) TaintSpotItn(nodeName string, eventID string) error {
 	return addTaint(k8sNode, n, SpotInterruptionTaint, eventID, corev1.TaintEffectNoSchedule)
 }
 
+// LogPods logs all the pod names on a node
+func (n Node) LogPods(nodeName string) error {
```
haugenj, line 291: Can you add a unit test or two for this within `node_test.go` ?

haugenj (changes requested):
Looks good, I've got two small requests before merging

pkg/node/node.go
haugenj:
```
func TestLogPods(t *testing.T) {
	resetFlagsForTest()

	client := fake.NewSimpleClientset(
		&v1.Pod{
			ObjectMeta: metav1.ObjectMeta{
				Name: "myPod",
				Labels: map[string]string{
					"spec.nodeName": nodeName,
				},
			},
		},
		&v1.Node{
			ObjectMeta: metav1.ObjectMeta{
				Name: nodeName,
			},
		},
	)

	tNode := getNode(t, getDrainHelper(client))
	err := tNode.LogPods(nodeName)
	h.Ok(t, err)
}
```

pkg/node/node.go
wanwenli: Sure. I have made the change.

pkg/node/node.go
wanwenli: Yes, test added. Thanks for your code snippet.

wanwenli:
FYI, @haugenj, the build of this PR has been blocked by `go-report-card-test`. To be more specific, [this line](https://github.com/aws/aws-node-termination-handler/blob/c9813c2796a98a9c94f6017769481ec52dc7e6eb/test/go-report-card-test/Dockerfile#L6) that should install `goreportcard`.

I test it on my laptop and in a docker container, and both cases fail with the same reasons.

The build on the main branch would face the same issue.

haugenj (approved):
👍 Thanks!

haugenj:
Looks like we're seeing the same issue with go report card in some of our other repos related to https://github.com/gojp/goreportcard/pull/298

Not sure what recently changed that's causing this to affect us now though, but regardless I don't think we should be held up on the report card. 

Thanks for the contribution!

PR #274, opened 2020-11-03
Add debug log messages to `isInstanceManaged`

kppullin:
This change adds debug log messages to give hints when an Auto Scaling
Group is not tagged as managed. Without this change there's no other
indication that the instance & ASG is unmanaged and makes diagnosing
this condition quite difficult without reading through the source code.

### DIFF
--- a/pkg/monitor/sqsevent/sqs-monitor.go
+++ b/pkg/monitor/sqsevent/sqs-monitor.go
@@ -194,6 +194,7 @@ func (m SQSMonitor) isInstanceManaged(instanceID string) (bool, error) {
 		return false, err
 	}
 	if len(asgs.AutoScalingInstances) == 0 {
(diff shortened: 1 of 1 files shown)


pkg/monitor/sqsevent/sqs-monitor.go
```diff
@@ -194,6 +194,7 @@ func (m SQSMonitor) isInstanceManaged(instanceID string) (bool, error) {
 		return false, err
 	}
 	if len(asgs.AutoScalingInstances) == 0 {
+		log.Debug().Str("instance_id", instanceID).Msg("Did not find an Auto Scaling Instance for the given instance id")
```
bwagner5, line 197: can you change "Auto Scaling instance" to "Auto Scaling Group"

kppullin: Done!

bwagner5 (approved):
lgtm thanks!

PR #280, opened 2020-11-06
Minor fixes

brycahta:
Issue #, if available:
N/A

Description of changes:
* AEMM and overriding instance metadata URL not needed for sqs tests --> removed
* Helm ReadMe zero-config instructs return an error for me (-- name does not exist) --> correcting it and now it aligns with main ReadMe

### DIFF
--- a/config/helm/aws-node-termination-handler/README.md
+++ b/config/helm/aws-node-termination-handler/README.md
@@ -16,8 +16,9 @@ Install AWS Node Termination Handler:
 To install the chart with the release name aws-node-termination-handler and default configuration:
 
 ```sh
-helm install --name aws-node-termination-handler \
-  --namespace kube-system eks/aws-node-termination-handler
+helm upgrade --install aws-node-termination-handler \
+  --namespace kube-system \
+  eks/aws-node-termination-handler
 ```
 
 To install into an EKS cluster where the Node Termination Handler is already installed, you can run:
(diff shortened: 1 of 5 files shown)


test/e2e/asg-lifecycle-sqs-test
```diff
@@ -40,22 +40,6 @@ set -x
 helm "${localstack_helm_args[@]}"
 set +x
 
-aemm_helm_args=(
```
bwagner5, line 43: We should fix the recent issue opened about the delay in starting up NTH if IMDS is not available because of the retires.  Removing this will slow down tests and might cause transient failures because of the assertion loop timeouts. I think it's the right move to remove AEMM from these test, just want to make sure the tests still pass consistently.

brycahta: on it like a bonnet 👒

bwagner5 (approved):
all the tests passed, so I'm cool with going forward with this

config/helm/aws-node-termination-handler/README.md
```diff
@@ -16,8 +16,9 @@ Install AWS Node Termination Handler:
 To install the chart with the release name aws-node-termination-handler and default configuration:
 
 ```sh
-helm install --name aws-node-termination-handler \
```
bwagner5, line 19: this is helm2 syntax FYI, but I agree helm3 all the way!

Issue #277, opened 2020-11-05
Support instance metadata skipping/bypass via configuration


backjo:
It would be useful to support bypassing the IMDS metadata scrape at startup. 

We're currently running the NTH Queue Processor, which consumes SQS events using IRSA-based credentials. Since this deployment is only relying on the queue events, we restrict it's access to IMDS via NetworkPolicy (global restriction in our cluster, with only a few select deployments allowed to access IMDS). 

Right now, it spends an unnecessary 40 seconds at startup trying to fetch IMDS data, because of 4 calls to this loop https://github.com/aws/aws-node-termination-handler/blob/5150b31c819a26c2cc80f812ffb0a0b317290b56/pkg/ec2metadata/ec2metadata.go#L191-L223, which tries 2 times  (hardcoded) per metadata call to fetch the metadata, waiting 5 seconds (hardcoded) per-call to timeout.

bwagner5:
Agreed, queue-processor mode should definitely be able to operate with IMDS disabled without significant startup delays. I can do some tests, but maybe just decreasing the retries on the startup requests would be enough.   I'm hesitant to add another flag to bypass, since there's already a ton of configuration and I think it's getting to be a little too much, but that may be the best solution.

IMDS is also useful in queue-processor mode for webhooks context. Are you using webhook notifications within NTH?

backjo:
We aren't today - potentially in the future though.

brycahta:
@backjo -- feel free to take a look at #281 which addresses this issue. When setting `metadataTries=0` I'm seeing startup time drop from ~40sec to ~2sec (more details in PR)

bwagner5:
released in https://github.com/aws/aws-node-termination-handler/releases/tag/v1.11.0

PR #281, opened 2020-11-09
reduce imds retries on error

brycahta:
Issue #, if available:
* #277 

Description of changes:
* Reduce connection timeout to `2 sec`
* If there are issues querying IMDS (specifically `/latest/dynamic/instance-identity/document`) on startup, then do **not** continue to query IMDS

Testing:
* Before changes, NTH would take 40+ sec to startup if calls to IMDS were failing
* Tested on local cluster:
  * `metadataTries=0` -->  2sec startup
  * `metadataTries=3` (default) -->  11sec startup


bwagner5 (approved):
nice! LGTM

PR #287, opened 2020-11-12
Allow the NTHManagedAsg tag to be configurable

### LINKED ISSUE #272: Failing to drain/cordon causes CrashLoopBackOff
[Queue processor specific]

Running into issue here around the [behavior of exiting with an error code of `1`](https://github.com/aws/aws-node-termination-handler/blob/5150b31c819a26c2cc80f812ffb0a0b317290b56/cmd/node-termination-handler.go#L260-L280) when a message is pulled from SQS and then cordoning or draining of that node fails.

In our use-case, we have multiple EKS clusters in an account in a region. When approaching how to handle this with NTH, I was going to run per-cluster SQS queues with CloudWatch events + rules that were tailored to the proper cluster/ASG combo, and this works for ASG events, but EC2 termination events do not have the same filtering capacity -- meaning that you end up with events in SQS queues even though those instances may not be in the cluster. When NTH processes one of those, it will `exit(1)` because there will be no node in the cluster to cordon or drain. After so many, Kubernetes will mark the pod in a `CrashLoopBackOff` state and you will lose all NTH capabilities until the cool-down period expires and Kubernetes starts the pod again.

For an NTH pod that I've been running for a week, it's seen 600+ crashes:
```
aws-node-termination-handler-86d9c789f4-6c9pz   0/1     CrashLoopBackOff     607        6d2h
```

```
2020/10/29 17:29:57 ??? Trying to get token from IMDSv2
2020/10/29 17:29:57 ??? Got token from IMDSv2
2020/10/29 17:29:57 ??? Startup Metadata Retrieved metadata={"accountId":"0xxxxxxxxx","availabilityZone":"us-east-1b","instanceId":"i-0xxxxxxxx","instanceType":"c5.2xlarge","localHostname":"ip-10-xxx-xxx-xxx.ec2.internal","privateIp":"10.xxx.xxx.xxx","publicHostname":"","publicIp":"","region":"us-east-1"}
2020/10/29 17:29:57 ??? aws-node-termination-handler arguments: 
	dry-run: false,
	node-name: ip-10-xxx-xxx-xxx.ec2.internal,
	metadata-url: http://169.254.169.254,
	kubernetes-service-host: 172.20.0.1,
	kubernetes-service-port: 443,
	delete-local-data: true,
	ignore-daemon-sets: true,
	pod-termination-grace-period: -1,
	node-termination-grace-period: 120,
	enable-scheduled-event-draining: false,
	enable-spot-interruption-draining: false,
	enable-sqs-termination-draining: true,
	metadata-tries: 3,
	cordon-only: false,
	taint-node: false,
	json-logging: false,
	log-level: INFO,
	webhook-proxy: ,
	webhook-headers: <not-displayed>,
	webhook-url: ,
	webhook-template: <not-displayed>,
	uptime-from-file: ,
	enable-prometheus-server: false,
	prometheus-server-port: 9092,
	aws-region: us-east-1,
	queue-url: https://sqs.us-east-1.amazonaws.com/0xxxxxxxxxx/node_termination_handler,
	check-asg-tag-before-draining: true,
	aws-endpoint: ,

2020/10/29 17:29:57 ??? Started watching for interruption events
2020/10/29 17:29:57 ??? Kubernetes AWS Node Termination Handler has started successfully!
2020/10/29 17:29:57 ??? Started watching for event cancellations
2020/10/29 17:29:57 ??? Started monitoring for events event_type=SQS_TERMINATE
2020/10/29 17:30:00 ??? Adding new event to the event store event={"Description":"Spot Interruption event received. Instance will be interrupted at 2020-10-29 17:28:15 +0000 UTC \n","Drained":false,"EndTime":"0001-01-01T00:00:00Z","EventID":"spot-itn-event-12345","InstanceID":"i-0xxxxxxxxxx","Kind":"SQS_TERMINATE","NodeName":"ip-10-xxx-xxx-xxx.ec2.internal","StartTime":"2020-10-29T17:28:15Z","State":""}
2020/10/29 17:30:01 ??? Cordoning the node
2020/10/29 17:30:01 WRN Error when trying to list Nodes w/ label, falling back to direct Get lookup of node
2020/10/29 17:30:01 ??? There was a problem while trying to cordon and drain the node error="nodes \"ip-10-xxx-xxx-xxx.ec2.internal\" not found"
```

blakestoddard:
Removes the `NTHManagedASG` tag constant in favor of a configurable `ManagedAsgTag` option to allow you to run NTH in queue mode across multiple clusters in an account. EC2 instance events don't have any filtering ability currently, which will cause NTH to drop into a crash loop when it runs into an error while trying to cordon a node that doesn't exist in the cluster. See [here](https://github.com/aws/aws-node-termination-handler/blob/5150b31c819a26c2cc80f812ffb0a0b317290b56/cmd/node-termination-handler.go#L260-L280).

Fixes #272

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -150,6 +150,7 @@ func main() {
 
 		sqsMonitor := sqsevent.SQSMonitor{
 			CheckIfManaged:   nthConfig.CheckASGTagBeforeDraining,
+			ManagedAsgTag:    nthConfig.ManagedAsgTag,
 			QueueURL:         nthConfig.QueueURL,
 			InterruptionChan: interruptionChan,
 			CancelChan:       cancelChan,
--- a/config/helm/aws-node-termination-handler/README.md
+++ b/config/helm/aws-node-termination-handler/README.md
@@ -85,6 +85,8 @@ Parameter | Description | Default
 `enableSqsTerminationDraining` | If true, this turns on queue-processor mode which drains nodes when an SQS termination event is received| `false`
 `queueURL` | Listens for messages on the specified SQS queue URL | None
 `awsRegion` | If specified, use the AWS region for AWS API calls, else NTH will try to find the region through AWS_REGION env var, IMDS, or the specified queue URL | ``
+`checkASGTagBeforeDraining` | If true, check that the instance is tagged with "aws-node-termination-handler/managed" as the key before draining the node | `true`
+`managedAsgTag` | The tag to ensure is on a node if checkASGTagBeforeDraining is true | `aws-node-termination-handler/managed`
 
 ### AWS Node Termination Handler - IMDS Mode Configuration
 
--- a/config/helm/aws-node-termination-handler/templates/deployment.yaml
+++ b/config/helm/aws-node-termination-handler/templates/deployment.yaml
@@ -140,6 +140,8 @@ spec:
           {{- end }}
           - name: CHECK_ASG_TAG_BEFORE_DRAINING
             value: {{ .Values.checkASGTagBeforeDraining | quote }}
+          - name: MANAGED_ASG_TAG
+            value: {{ .Values.managedAsgTag | quote }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           {{- if .Values.enablePrometheusServer }}
--- a/config/helm/aws-node-termination-handler/values.yaml
+++ b/config/helm/aws-node-termination-handler/values.yaml
@@ -45,6 +45,9 @@ queueURL: ""
 # checkASGTagBeforeDraining  If true, check that the instance is tagged with "aws-node-termination-handler/managed" as the key before draining the node
 checkASGTagBeforeDraining: true
 
+# managedAsgTag  The tag to ensure is on a node if checkASGTagBeforeDraining is true
+managedAsgTag: "aws-node-termination-handler/managed"
+
 # awsRegion If specified, use the AWS region for AWS API calls
 awsRegion: ""
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -290,6 +290,7 @@ func (c Config) PrintHumanConfigArgs() {
 			"\taws-region: %s,\n"+
 			"\tqueue-url: %s,\n"+
 			"\tcheck-asg-tag-before-draining: %t,\n"+
+			"\tmanaged-asg-tag: %s,\n"+
 			"\taws-endpoint: %s,\n",
 		c.DryRun,
 		c.NodeName,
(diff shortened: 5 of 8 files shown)


brycahta (changes requested):
Looks like build is failing due to unit tests:

* `pkg/monitor/sqsevent/sqs-monitor_internal_test.go:51: undefined: NTHManagedASG`
* `pkg/monitor/sqsevent/sqs-monitor_test.go:542: undefined: NTHManagedASG`

blakestoddard:
I wasn't sure the best way to really fix that because passing a string to `mockIsManagedTrue` during struct initialization feels just as gross as just updating the constant reference in the function with a new string, so that's what I did.

brycahta (changes requested):
Looks like **unit tests** are still failing. After some deep dive:

* Looks like you need to initialize [SQSMonitor](https://github.com/aws/aws-node-termination-handler/blob/main/pkg/monitor/sqsevent/sqs-monitor_internal_test.go#L55) with a `ManagedAsgTag` set to the default/tag value; otherwise the `if *tag.Key == m.ManagedAsgTag` fails

brycahta (changes requested):
Could you also update the **Helm artifacts** in this [folder](https://github.com/aws/aws-node-termination-handler/tree/main/config/helm/aws-node-termination-handler) including ReadMe so that `ManagedAsgTag` can also be configurable as a Helm param? 

the files should be:
* [values.yaml](https://github.com/aws/aws-node-termination-handler/blob/main/config/helm/aws-node-termination-handler/values.yaml)
* [deployment.yaml](https://github.com/aws/aws-node-termination-handler/blob/main/config/helm/aws-node-termination-handler/templates/deployment.yaml)
* [ReadMe](https://github.com/aws/aws-node-termination-handler/tree/main/config/helm/aws-node-termination-handler#aws-node-termination-handler-common-configuration)

edit: removing daemonset configs

blakestoddard:
I don't see any of the related queue processor variables (`queueURL`, `checkASGTagBeforeDraining`, etc.) defined in the two daemonset configs, so I didn't add those. Should they be?

brycahta:
> I don't see any of the related queue processor variables (`queueURL`, `checkASGTagBeforeDraining`, etc.) defined in the two daemonset configs, so I didn't add those. Should they be?

Ahh sorry that's my mistake. queue-processor vars shouldn't be in daemonset configs at all; i'll remove from comment

blakestoddard:
Are those failures in Travis related? Seems like they're from `prometheus-operator`.

brycahta:
> Are those failures in Travis related? Seems like they're from `prometheus-operator`.

Correct-- these failures aren't related to your changes.

brycahta (approved):
LGTM 👍🏼

bwagner5 (approved):
lgtm, thanks!

PR #288, opened 2020-11-12
🥑🤖 v1.11.0 release prep

brycahta:
🥑🤖 Auto-generated PR for v1.11.0 release. Updating release versions in repo.

### DIFF
--- a/config/helm/aws-node-termination-handler/Chart.yaml
+++ b/config/helm/aws-node-termination-handler/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 name: aws-node-termination-handler
 description: A Helm chart for the AWS Node Termination Handler
 version: 0.12.0
-appVersion: 1.10.0
+appVersion: 1.11.0
 home: https://github.com/aws/eks-charts
 icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
 sources:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -89,7 +89,7 @@ The termination handler DaemonSet installs into your cluster a [ServiceAccount](
 
 You can use kubectl to directly add all of the above resources with the default configuration into your cluster.
 ```
-kubectl apply -f https://github.com/aws/aws-node-termination-handler/releases/download/v1.10.0/all-resources.yaml
+kubectl apply -f https://github.com/aws/aws-node-termination-handler/releases/download/v1.11.0/all-resources.yaml
```
bwagner5, line 92: Can we leave off the readme update of the release automation?  Since NTH takes a while to release, at least 30-45 minutes, the readme will result in errors during that time. 

I think it's better to add a post-release step executed in Travis that uses ec2-bot to submit a PR back to the repo to update the README once all the assets are released (should be similar to the eks-charts sync script).

brycahta: Not opposed, but are we good with source code for a given release not having accurate version numbers in its ReadMe? Or were you thinking fewer users will be affected in this scenario vs. pulling a version that hasn't released yet

bwagner5: Yeah, I think that including the readme w/ previous version in the archive files attached to the release is fine. We don't tell anyone to use those archives and I don't see a reason, even if they did, to use that readme as a source of truth.

brycahta: will submit a separate PR to update as a post-release step

bwagner5 (approved):
👍

PR #312, opened 2020-12-09
add log-level helm param

brycahta:
Issue #, if available:
* N/A: helm param for log level was missing

Description of changes:
* added param

Testing:
* added various log levels (debug, info, error) to e2e tests, executed locally, and saw pod logs reflected the configuration

### DIFF
--- a/config/helm/aws-node-termination-handler/values.yaml
+++ b/config/helm/aws-node-termination-handler/values.yaml
@@ -76,6 +76,9 @@ taintNode: false
 # Log messages in JSON format.
 jsonLogging: false
 
+# Sets the log level
+logLevel: "info"
+
 # dryRun tells node-termination-handler to only log calls to kubernetes control plane
 dryRun: false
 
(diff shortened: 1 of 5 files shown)


config/helm/aws-node-termination-handler/values.yaml
```diff
@@ -76,6 +76,9 @@ taintNode: false
 # Log messages in JSON format.
 jsonLogging: false
 
+# Sets the log level
+logLevel: "info"
```
haugenj, line 80: should we leave this as `""` ?

brycahta: My thought is it's easier to figure out the default value this way.

 If we leave it as `""` , then helm readme would have default value say `""` too (even though app will default to 'info') OR we have the helm default value actually say `info`, but then the value in values.yaml would be `""` and not match which is confusing/doesn't align with how we documented other vals.

haugenj (approved):
👍

Issue #307, opened 2020-11-27
Multiple events for the same node cause a crash


Niksko:
Currently I'm sending both EC2 state change events and ASG lifecycle events to the NTH in queue processor mode. This sounds like it should be supported, however in practice what seems to be happening is:

1. ASG lifecycle event is handle. Node is cordoned and drained.
2. EC2 state change event for termination comes though, node is now no longer there, so the NTH crashes.

I'm not convinced that the NTH _needs_ to crash if it can't find a node, but that aside, is there any other workaround for this? Or should I just pick EC2 state change events OR ASG lifecycle events to handle?

bwagner5:
We've discussed a related situation previously here https://github.com/aws/aws-node-termination-handler/issues/272 

I think this issue https://github.com/aws/aws-node-termination-handler/issues/297 could be used to mitigate this situation but we may not be able to run multiple replicated NTH pods reliably. 

I'd be interested to see if using ec2 instance tags to turn on and off NTH management could mitigate this (although I'm not sure that would be the right use of it or work in every instance).

I'm not opposed to removing the crash and just log the event if it's not found. I don't think there's a possible workaround at this time.

Niksko:
I'd prefer the crash removal if possible. I can understand why not being able to find a node might have been worthy of causing a crash, but it seems increasingly that this is a common scenario.

universam1:
Agree facing same issue - in our case we basically see the pod crashing all the time

paalkr:
I still observe this issue with 1.11 if I enable both ASG lifecycle events
```
      EventPattern:
        source:
        - "aws.autoscaling"
        detail-type:
        - "EC2 Instance-terminate Lifecycle Action"
```

and EC2 state change events
```
      EventPattern:
        source:
        - "aws.ec2"
        detail-type:
        - "EC2 Instance State-change Notification"
        detail:
          state:
          - "shutting-down"
```

universam1:
@haugenj anything we can do to get a fix? Would you accept a PR?

haugenj:
PRs are always welcome!

If you could include a test to validate this scenario that would be 💯

universam1:
@haugenj fixed this issue by #313 

Background: went all the way back, instead of just removing the os.exit(1) found that actually the nodeName was empty but not verified.
The existing unit test was flawed in this regard.

[...]

PR #313, opened 2020-12-10
🐛 fixing crashes on State-change events

universam1:
NTH crashes often from "EC2 Instance State-change Notification" events.

fixes https://github.com/aws/aws-node-termination-handler/issues/307

The underlying issue was that an empty parsed NodeName derived from PrivateDnsName for forwarded unverified, creating cascading problems.

This solves the root cause, that "EC2 Instance State-change Notification" can arrive at a time where the instance is in shutting-down, terminated or any other not-online situation where the PrivateDnsName metadata is empty!

Instead of just ignoring these errors this implementation gets and decides based on the state of the instance message if it is an error to fail or to ignore.
Therefor such messages are dropped in above situation because they are useless.

@haugenj 

Background: went all the way back, instead of just removing the os.exit(1) found that actually the nodeName was empty but not verified.
The existing unit test was flawed in this regard.

The empty nodeName parsing derives from `EC2.DescribeInstances`  that have the metadata of `PrivateDnsName` empty, simply because the node is not running any more.

So created a custom error that allows to handle that case.

example event:
```json
{
    "AmiLaunchIndex": 1,
    "Architecture": "x86_64",

    "ImageId": "ami-0e44cbb6cd97xxxx",
    "InstanceId": "i-xxxxxxxx",

    "Platform": null,
    "PrivateDnsName": "",
    "PrivateIpAddress": null,
    "ProductCodes": null,
    "PublicDnsName": "",
    "PublicIpAddress": null,

    "State": {
        "Code": 48,
        "Name": "terminated"
    },
    "StateReason": {
        "Code": "Client.UserInitiatedShutdown",
        "Message": "Client.UserInitiatedShutdown: User initiated shutdown"
    },

}
```

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -37,6 +37,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/sqs"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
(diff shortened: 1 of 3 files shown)


dhohengassner:
nice - think I have seen such crashes already

would be happy if this fix gets merged :+1:

pkg/monitor/sqsevent/sqs-monitor_test.go
```diff
@@ -377,7 +377,7 @@ func TestMonitor_EC2NoInstances(t *testing.T) {
 	}
 }
 
-func TestMonitor_EC2NoDNSName(t *testing.T) {
+func TestMonitor_EC2NoDNSNameFailure(t *testing.T) {
```
haugenj, line 380: Not sure we should change the name here since the result is no error. With "Failure" in the name I would expect `err != nil`

pkg/monitor/sqsevent/sqs-monitor.go
```diff
@@ -33,6 +34,9 @@ const (
 	SQSTerminateKind = "SQS_TERMINATE"
 )
 
+// ErrNodeTerminated forwards condition that the instance is terminated thus metadata missing
```
haugenj, line 37: should `ErrNodeTerminated` be `ErrNodeMetadataUnavailable` ? 

What do you think about renaming this something more specific to the node state not being "running" since that's what we're explicitly checking on line 198? Maybe `ErrNodeStateNotRunning` or `ErrNodeNotRunning` ?

haugenj:
@universam1 This is great! Left you two small nitpick comments but this is really awesome, thanks!

pkg/monitor/sqsevent/sqs-monitor_test.go
universam1: fine with me

pkg/monitor/sqsevent/sqs-monitor.go
universam1: good point  - adjusted!

cmd/node-termination-handler.go
```diff
@@ -267,6 +268,11 @@ func drainOrCordonIfNecessary(interruptionEventStore *interruptioneventstore.Sto
 		if nthConfig.CordonOnly || drainEvent.IsRebalanceRecommendation() {
 			err := node.Cordon(nodeName)
 			if err != nil {
+				if errors.IsNotFound(err) {
```
universam1, line 271: @haugenj found another situation where events for other nodes that do not belong to this cluster could bubble up an error - I think for this case we can only do best effort to ignore such cases without removing all the safety guards

haugenj: I think this is the right approach. Originally I was thinking we would totally remove the `os.Exit(1)` here, so your solution feels like a better middle-ground

haugenj (approved):
👍  LGTM. Thanks!

PR #314, opened 2020-12-10
Fix duplicate error equality check

haugenj:
Issue #, if available:
no issue

Description of changes:
We're comparing the error objects instead of their contained errors. `err` is new on each loop so it'll never be equal to `previousErr`

Proof is in the playground: https://play.golang.org/p/o5ezN5xvaSy


brycahta (approved):
Nice catch 👍

PR #335, opened 2021-01-04
fix update version check

brycahta:
Issue #, if available:
* Last release failed because script prematurely exited when grep failed
  * https://github.com/aws/aws-node-termination-handler/runs/1602759774?check_suite_focus=true

Description of changes:
* Do not exit when grep is unable to find the previous version in documents, instead script should keep going and eventually have a file list of 0 requiring updates (and therefore no action)
* all other errors (exit code > 1) from grep will lead to script terminating


bwagner5 (approved):
LGTM

Issue #361, opened 2021-02-11
Node on "degraded hardware" not evacuated & terminated ASAP


tyrken:
I took advantage of the rare incidence of a scheduled event on a node in our test EKS cluster to update to the latest NTH (1.12.0 via latest helm chart), running in IMDS mode still.  Unfortunately it didn't start draining, despite getting the below notice I was able to capture manually:

```bash
$ curl http://169.254.169.254/latest/meta-data/events/maintenance/scheduled
[ {
  "NotBefore" : "25 Feb 2021 02:00:00 GMT",
  "Code" : "instance-stop",
  "Description" : "The instance is running on degraded hardware",
  "EventId" : "instance-event-087e363178449a61e",
  "State" : "active"
} 
```

helm settings (set via helmfile but I suspect you understand):
```yaml
repositories:
  - name: aws-eks-charts
    url: https://aws.github.io/eks-charts
releases:
  - name: aws-node-termination-handler
    namespace: kube-system
    chart: aws-eks-charts/aws-node-termination-handler
    version: 0.13.3
    values:
      - enableScheduledEventDraining: true
        enableSpotInterruptionDraining: true
        enableRebalanceMonitoring: false
        taintNode: true
        webhookURL: <redacted>
        enablePrometheusServer: true
        nodeSelector:
          lifecycle: spot
        podMonitor:
          create: true
        jsonLogging: true
        logLevel: DEBUG
```

deamonset pod logs:
```
{"time":"2021-02-11T09:35:48Z","message":"Trying to get token from IMDSv2"}
{"level":"info","time":"2021-02-11T09:35:48Z","message":"Starting to serve handler /metrics, port 9092"}
{"time":"2021-02-11T09:35:48Z","message":"Got token from IMDSv2"}
{"metadata":{"accountId":"XXXXXXXX","instanceId":"i-0415e8c43a18dccc7","instanceType":"t2.medium","publicHostname":"ec2-54-229-104-67.eu-west-1.compute.amazonaws.com","publicIp":"54.229.104.67","localHostname":"ip-172-31-13-215.empire.lyst","privateIp":"172.31.13.215","availabilityZone":"eu-west-1b","region":"eu-west-1"},"time":"2021-02-11T09:35:48Z","message":"Startup Metadata Retrieved"}
[...]

Issue #315, opened 2020-12-10
Unable to parse scheduled event end time


mikesplain:
Hello! I was just looking at one of our instances running the aws-node-termination-handler after getting a ec2 degredation email about it.  I was curious if NTH had removed the instances since we have `ENABLE_SCHEDULED_EVENT_DRAINING` enabled.  

We are running `amazon/aws-node-termination-handler:v1.11.0` and  we're using kiam underneath for IAM access.  

```json
{"time":"2020-12-10T00:59:27Z","message":"Trying to get token from IMDSv2"}
{"time":"2020-12-10T00:59:27Z","message":"Got token from IMDSv2"}
{"time":"2020-12-10T00:59:27Z","message":"Trying to get token from IMDSv2"}
{"time":"2020-12-10T00:59:27Z","message":"Got token from IMDSv2"}
{"error":"parsing time \"\" as \"2 Jan 2006 15:04:05 GMT\": cannot parse \"\" as \"2\"","time":"2020-12-10T01:50:29Z","message":"Unable to parse scheduled event end time, continuing"}
{"event":{"EventID":"instance-event-0c6a1ed93bfc96fd4","Kind":"SCHEDULED_EVENT","Description":"instance-stop will occur between 28 Dec 2020 06:00:00 GMT and  because The instance is running on degraded hardware\n","State":"active","NodeName":"ip-10-223-202-136.ec2.internal","InstanceID":"","StartTime":"2020-12-28T06:00:00Z","EndTime":"2020-12-28T06:00:00Z","Drained":false},"time":"2020-12-10T01:50:29Z","message":"Adding new event to the event store"}
{"error":"parsing time \"\" as \"2 Jan 2006 15:04:05 GMT\": cannot parse \"\" as \"2\"","time":"2020-12-10T01:50:31Z","message":"Unable to parse scheduled event end time, continuing"}
{"error":"parsing time \"\" as \"2 Jan 2006 15:04:05 GMT\": cannot parse \"\" as \"2\"","time":"2020-12-10T01:50:33Z","message":"Unable to parse scheduled event end time, continuing"}
# This continues like this indefinitely, even after a pod restart
```

Let me know if any more info would be helpful!

haugenj:
NTH doesn't actually terminate instances, it'll just drain and/or cordon them. To see if NTH handled this event correctly you can check the node to see if it's been tainted with `aws-node-termination-handler/scheduled-maintenance`.

[...]

PR #363, opened 2021-02-12
Avoid noisy error on scheduled events without NotAfter

tyrken:
See Issue #361 and related #315

Description of changes:
Hide the continual error from attempting to parse an empty string (when there isn't a `NotAfter` in the scheduled event, as in the example from #361 ).

Note this only fixes the minor point of the issue - that the node isn't immediately drained & eventually terminated remains.


haugenj (approved):
👍  LGTM, thanks!

PR #364, opened 2021-02-12
Sqs deletion issue 360

### LINKED ISSUE #360: SQS messages deleted without being processed
When several nodes are being terminated at once, we've observed that some of the messages are deleted from the queue without being processed.

This seems to be due to NTH retrieving [up to two messages from SQS](https://github.com/aws/aws-node-termination-handler/blob/23b97f40987f52a28d533bd7bba7a391d46e7e3b/pkg/monitor/sqsevent/sqs-monitor.go#L143), [only processing the first](https://github.com/aws/aws-node-termination-handler/blob/23b97f40987f52a28d533bd7bba7a391d46e7e3b/pkg/monitor/sqsevent/sqs-monitor.go#L78) but passing both into `asgTerminationToInterruptionEvent` where [both messages are deleted](https://github.com/aws/aws-node-termination-handler/blob/23b97f40987f52a28d533bd7bba7a391d46e7e3b/pkg/monitor/sqsevent/asg-lifecycle-event.go#L114).

azenk:
Issue #360

Fixes #360 by iterating over all messages retrieved from SQS.  The tests are also updated to process all three test events at once without spawning a go process.

I have at least one open question: 

*Should further processing of messages from the queue continue if an error is encountered?*
Currently those events are silently dropped for future retry.  SQS will bury the unprocessed events until the visibility timeout expires and then redeliver them.  A few possible options:
* Just let them get ignored.  This slows down handling of events, and is likely sub-optimal
* Continue processing the next message in the queue, but that changes the assumed semantics of Monitor() a bit, as errors couldn't be surfaced unless they're merged together.
* Forget processing multiple messages per receive.  Set the max sqs receive count to 1, which would allow it to only bury messages that generate errors and retain existing semantics at the cost of a few extra API calls under load.

### DIFF
--- a/pkg/monitor/sqsevent/asg-lifecycle-event.go
+++ b/pkg/monitor/sqsevent/asg-lifecycle-event.go
@@ -57,7 +57,7 @@ type LifecycleDetail struct {
 	LifecycleTransition  string `json:"LifecycleTransition"`
 }
 
-func (m SQSMonitor) asgTerminationToInterruptionEvent(event EventBridgeEvent, messages []*sqs.Message) (monitor.InterruptionEvent, error) {
+func (m SQSMonitor) asgTerminationToInterruptionEvent(event EventBridgeEvent, message *sqs.Message) (monitor.InterruptionEvent, error) {
 	lifecycleDetail := &LifecycleDetail{}
 	err := json.Unmarshal(event.Detail, lifecycleDetail)
 	if err != nil {
@@ -94,7 +94,7 @@ func (m SQSMonitor) asgTerminationToInterruptionEvent(event EventBridgeEvent, me
 		log.Info().Msgf("Completed ASG Lifecycle Hook (%s) for instance %s",
 			lifecycleDetail.LifecycleHookName,
 			lifecycleDetail.EC2InstanceID)
-		errs := m.deleteMessages(messages)
+		errs := m.deleteMessages([]*sqs.Message{message})
 		if errs != nil {
 			return errs[0]
 		}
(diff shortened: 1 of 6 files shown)


haugenj:
Thanks for writing this and for your patience in getting a response. We're based out of central Texas and are dealing with a bad winter storm right now. 

The intention with returning those errors is that we can kill NTH if the same error keeps occurring, the idea being something has been misconfigured and NTH can't function properly. 

With your change we still cover errors in receiving messages, it's just the processing of messages that is trickier to handle now. I see you updated the PR to ignore the errors in that case, which I think is a fine solution. 

Another idea is that we could perform a duplicate error check within sqs-monitor.go, and panic from within that file if we get too many consecutive duplicate errors. Essentially, copying the panic functionality from `node-termination-handler.go` into `sqs-monitor.go`. I don't think this is necessary, but am curious to hear your thoughts

azenk:
Based on the suggestion that @haugenj made to follow the pattern used for scheduled events, I opted to simply continue processing the next message in the queue after logging any errors.  This makes the current testing a little less than ideal.  As things stand right now, most errors aren't actually checked.  The tests simply verify that `Monitor()` doesn't error and no events are generated.

Much of the error testing should probably be pushed into the internal tests running directly against `processSQSMessage()` or the various event building functions.  I'd like to do this work, but it's a larger effort than I have time for at the moment.

As an aside, have you considered using something like [testify](https://github.com/stretchr/testify) for assertions?  I found the current helper library a bit confusing.  The testify output is also very clean and easy to read.

azenk:
> Another idea is that we could perform a duplicate error check within sqs-monitor.go, and panic from within that file if we get too many consecutive duplicate errors. Essentially, copying the panic functionality from `node-termination-handler.go` into `sqs-monitor.go`. I don't think this is necessary, but am curious to hear your thoughts

How about returning something like `AllMessageProcessingFailed` if we can't process any of the messages in the queue?  That should prevent us from getting stuck forever since bad messages eventually get removed.  If the duplicate error threshold is above the SQS delivery retry maximum, a single bad message should never cause an issue.  Perhaps that should be documented if we go that route?

haugenj:
That sounds good to me. I think we might as well also increase the number of SQS messages taken in a single receive request now that we can properly handle them. Perhaps up from 2 to 5 now?

haugenj (approved):
👍  LGTM. Thanks for going the extra mile improving the tests!

Issue #346, opened 2021-01-13
Queue process docs don't describe non-Helm config


gabegorelick:
The docs for setting up NTH in queue processor mode don't make it clear that patches must be applied to the YAML to set queue URL and enableSqsTerminationDraining. You have to read the Helm section further down to realize that these settings must be applied, and then figure out where in the manifests those changes should be made.

![image](https://user-images.githubusercontent.com/610129/104500294-3e20b980-55ac-11eb-83d3-c51963505cc7.png)

I think the "You can use kubectl to directly add all of the above resources with the default configuration into your cluster" section was copied from the IMDS docs. But the difference is that IMDS mode doesn't require any config.

bwagner5:
that's a good point.  Are you planning on consuming NTH from the manifest files in the release assets? 
If so, what transform tool are you using?  We used to have instructions for kustomize, maybe it's time to bring that back for this configuration.

Most people are consuming NTH w/ helm.

gabegorelick:
> Are you planning on consuming NTH from the manifest files in the release assets?

Yep. 

> If so, what transform tool are you using?

It's complicated. 😄 It's a mix of kustomize and custom templating to fill in the gaps. 

> We used to have instructions for kustomize, maybe it's time to bring that back for this configuration.

Since `kubectl apply -k` has become somewhat of a standard, I think it makes sense to document that.

> Most people are consuming NTH w/ helm.

You may want to consider making helm the preferred method of installation in the docs (moving it above `kubectl apply`, maybe adding a sentence about how it's most commonly used). Especially since all the config is only documented in the helm section of the codebase (which is a separate issue).

PR #366, opened 2021-02-16
clarify queue processor install instructions

brycahta:
Issue #, if available: #346

Description of changes:
* Moving Helm install instructions above kubectl since it is the most popular installation method and clearest in terms of configuring NTH
* Added instructions for the minimal config necessary when using kubectl. Noted that the default configs alone will not work because sqs queue will be ""


haugenj (approved):
👍 LGTM

PR #369, opened 2021-02-19
Fix goreportcard test and run make fmt

haugenj:
Description of changes:
golang version 1.16 was released this week, and they changed the default behavior of `GO111MODULE`. So now if we want it to act like it used to, we have to set `GO111MODULE=auto`. More info on `GO111MODULE` [here](https://golang.org/ref/mod#mod-commands)

It was not downloading the source files for goreportcard, so when trying to `cd` into that directory the directory did not exist.

I also ran `make fmt`, which changed one line in `sqs-monitor_test.go` 


bwagner5 (approved):
lgtm

Issue #365, opened 2021-02-15
Missing permission in the managed podSecurityPolicy to authorize the pod to bind port 9092 in hostNetwork mode


mvisonneau:
When attempting to use `enablePrometheusServer: true` in conjunction with `rbac.pspEnabled: true` and `useHostNetwork: true`. It looks like there is some missing parameter in the PodSecurityPolicy to authorize the pod to bind `TCP/9092`.

```
Warning  FailedCreate      69s (x9 over 3m54s)    daemonset-controller  Error creating: pods "aws-node-termination-handler-" is forbidden: unable to validate against any pod security policy: [spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used spec.volumes[0]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.containers[0].hostPort: Invalid value: 9092: Host port 9092 is not allowed to be used. Allowed ports: [] spec.containers[0].hostPort: Invalid value: 9092: Host port 9092 is not allowed to be used. Allowed ports: []]
```

haugenj:
How did you install NTH? I've just tried reproducing this on a cluster of mine and haven't had any errors. 

If you edit the pod security policy and add the following is there any difference? 
```
hostPorts:
  - min: 9092
    max: 9092
```

If that fixes your issue we can change the yaml template

mvisonneau:
👋 hey @haugenj, absolutely this is exactly what I did and it sorts the issue out 👍

I can submit a PR with the change if you want me to, I just did not take the time to do it sorry about that! 😅

haugenj:
Yeah if you could write it that'd be awesome! If you can use Mustache to dynamically get the value of the Prometheus port and only include it if Prometheus is enabled I think that would be best. Thanks!

PR #371, opened 2021-02-21
helm: added necessary hostPorts in the PodSecurityPolicy

mvisonneau:
When metric/prometheus endpoint is enabled, we currently have an issue as the
PodSecurityPolicy does not allow the pod to bind the ports. This change sorts
this issue out (#365).

cc @haugenj


haugenj (approved):
LGTM, thanks!

PR #372, opened 2021-02-24
helm: Add support to override pod monitor namespace

victorboissiere:
# Description of changes

The pod monitor is only created in the Helm release namespace. The Prometheus operator by default watches pod monitors resources only in its own namespace. To watch resources in another namespace, it can be configured to use label selectors. However, in some cases, we do not want to add labels to an already existing namespace like kube-system.

This PR allows overriding the release namespace for the pod monitor. Some other Helm charts already support this feature like the pushgateway.

# Testing
Forked the Helm chart and deployed it. Working as intended.
helm template output is working as well: previous release namespace and overridden namespace when the added namespace is uncommented.


bwagner5 (approved):
lgtm! Thanks!

Issue #367, opened 2021-02-17
panic: No instance found with instance-id


ichasco-heytrade:
Hi, I have configured NTH with SQS and I am getting errors when a spot instance is released. It still detecting it in SQS (I think) and giving errors because now it doesn't exists.

```
...
2021/02/17 19:25:16 DBG Checking for queue messages
2021/02/17 19:25:16 ??? There was a problem monitoring for events error="No instance found with instance-id i-036d0327276ea8741" event_type=SQS_TERMINATE
2021/02/17 19:25:18 DBG Checking for queue messages
...
2021/02/17 19:25:54 ??? There was a problem monitoring for events error="No instance found with instance-id i-036d0327276ea8741" event_type=SQS_TERMINATE
2021/02/17 19:25:54 ??? Stopping NTH - Duplicate Error Threshold hit.
panic: No instance found with instance-id i-036d0327276ea8741

goroutine 72 [running]:
main.main.func3(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1ca8720, ...)
	/node-termination-handler/cmd/node-termination-handler.go:184 +0x4d1
created by main.main
	/node-termination-handler/cmd/node-termination-handler.go:167 +0x91c
 ``` 
 
 Version of NTH: `v1.12.0`

Configuration:

```
2021/02/17 19:31:17 ??? Startup Metadata Retrieved metadata={"accountId":"XXXXXXXX","availabilityZone":"eu-west-1b","instanceId":"i-090bb5ba647dad92e","instanceType":"m5.xlarge","localHostname":"ip-10-11-10-152","privateIp":"10.11.10.152","publicHostname":"","publicIp":"","region":"eu-west-1"}
2021/02/17 19:31:17 ??? aws-node-termination-handler arguments: 
	dry-run: false,
	node-name: ip-10-11-10-152.eu-west-1.compute.internal,
	metadata-url: http://169.254.169.254,
	kubernetes-service-host: 172.20.0.1,
	kubernetes-service-port: 443,
	delete-local-data: true,
	ignore-daemon-sets: true,
	pod-termination-grace-period: -1,
	node-termination-grace-period: 120,
	enable-scheduled-event-draining: false,
	enable-spot-interruption-draining: false,
	enable-sqs-termination-draining: true,
	enable-rebalance-monitoring: false,
	metadata-tries: 3,
	cordon-only: false,
	taint-node: false,
	json-logging: false,
[...]

PR #375, opened 2021-03-01
Drop sqs messages for absent nodes

haugenj:
Issue #367, if available:

Description of changes:
If no instances are returned from the describe instances call then there's no work to do, so we'll return a `ErrNodeStateNotRunning` error. This gets handled in the `Monitor` function - it'll delete the message and avoid the panic. 

With this we can also remove the `if nodeName == "" {` check in asg-lifecycle-event because we never actually run that. There's always a nodename by that point or we'll have already returned an error. 


bwagner5 (approved):
lgtm!

Issue #368, opened 2021-02-18
Feature request: Drain on Rebalance Recommendation Notification 


dongheeJeong:
Spot interruption termination Notification occurs in two minutes before instance termination. Two minutes may not sufficient time to drain node on some circumstance. (e.g. When new worker node creation needed when all existing worker nodes are too busy to accept pods). 

If option to drain on Rebalance Recommendation Notification(which arrives sooner than interruption termination Notification) provided, it would offer more time to pod to move to the other node.

Is there a plan on the roadmap to support something like this?

haugenj:
We have vague plans of doing a refactor at some point to allow for better customization for any event type, this would help in other cases too such as #361 for scheduled events. But for now draining on a rebalance recommendation is not something we plan on supporting in a configuration option. 

I'm not sure I understand your situation, do you mean the node is unable to completely drain within two minutes? Unless the drain is failing to complete I don't see how draining sooner than the spot itn will increase the speed of provisioning a new node

dongheeJeong:
I use [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) which add new worker node when there are pods that failed to run in the cluster due to insufficient resources. When drain needed and if there are insufficient resources in cluster for moving pods, new Worker node would provisioned (by Cluster Autoscaler) **before starting drain**. So, it takes time of `new node provisioning` + `drain`.

I do not mean that time of `new node provisioning` + `drain` is always more than 2 minutes, but it could take longer than general situation.

opalmer:
> But for now draining on a rebalance recommendation is not something we plan on supporting in a configuration option.

@haugenj, would you be open to an MR to add this or would you consider supporting this feature? 

Right now, based on what I've observed in a couple of clusters, when using the spot termination handler to handle rebalancing notifications the handler picks up the event and cordons the node. It does not drain the node to remove pods. In this scenario the next thing the ASG does is launch a new instance and then finally terminate the instance that got the rebalance recommendation. The AWS termination handler _does not_ drain the node as a result because it never receives a "spot instance interruption" signal at least based on the logs I've seen from the service. Essentially, operating in this mode currently is not much better than running without the spot termination handler because it it causes workloads to be left on the node until it's terminated.

haugenj:
👋 @opalmer 

[...]

PR #400, opened 2021-04-07
feat: add ability to drain on rebalance

trutx:
Issue #, if available: https://github.com/aws/aws-node-termination-handler/issues/368

Description of changes:

Currently NTH will only cordon a node when a rebalance recommendation signal is received. This PR adds the ability to also drain the node when such signal is received.

Also adding a couple of missing parameters in the Helm chart.

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -145,7 +145,7 @@ func main() {
 		imdsScheduledEventMonitor := scheduledevent.NewScheduledEventMonitor(imds, interruptionChan, cancelChan, nthConfig.NodeName)
 		monitoringFns[scheduledMaintenance] = imdsScheduledEventMonitor
 	}
-	if nthConfig.EnableRebalanceMonitoring {
+	if nthConfig.EnableRebalanceMonitoring || nthConfig.EnableRebalanceDraining {
 		imdsRebalanceMonitor := rebalancerecommendation.NewRebalanceRecommendationMonitor(imds, interruptionChan, nthConfig.NodeName)
 		monitoringFns[rebalanceRecommendation] = imdsRebalanceMonitor
 	}
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -291,7 +291,7 @@ func drainOrCordonIfNecessary(interruptionEventStore *interruptioneventstore.Sto
 		metrics.NodeActionsInc("pre-drain", nodeName, err)
 	}
 
-	if nthConfig.CordonOnly || drainEvent.IsRebalanceRecommendation() && !nthConfig.DrainOnRebalance {
+	if nthConfig.CordonOnly || (drainEvent.IsRebalanceRecommendation() && !nthConfig.EnableRebalanceDraining) {
 		err := node.Cordon(nodeName)
 		if err != nil {
 			if errors.IsNotFound(err) {
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -291,7 +291,7 @@ func drainOrCordonIfNecessary(interruptionEventStore *interruptioneventstore.Sto
 		metrics.NodeActionsInc("pre-drain", nodeName, err)
 	}
 
-	if nthConfig.CordonOnly || drainEvent.IsRebalanceRecommendation() {
+	if nthConfig.CordonOnly || drainEvent.IsRebalanceRecommendation() && !nthConfig.DrainOnRebalance {
 		err := node.Cordon(nodeName)
 		if err != nil {
 			if errors.IsNotFound(err) {
--- a/config/helm/aws-node-termination-handler/Chart.yaml
+++ b/config/helm/aws-node-termination-handler/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: aws-node-termination-handler
 description: A Helm chart for the AWS Node Termination Handler
-version: 0.15.1
+version: 0.14.2
 appVersion: 1.12.2
 home: https://github.com/aws/eks-charts
 icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
--- a/config/helm/aws-node-termination-handler/Chart.yaml
+++ b/config/helm/aws-node-termination-handler/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: aws-node-termination-handler
 description: A Helm chart for the AWS Node Termination Handler
-version: 0.14.2
+version: 0.15.1
 appVersion: 1.12.2
 home: https://github.com/aws/eks-charts
 icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
--- a/config/helm/aws-node-termination-handler/README.md
+++ b/config/helm/aws-node-termination-handler/README.md
@@ -69,7 +69,8 @@ Parameter | Description | Default
 `webhookTemplateConfigMapKey` | Name of the template file stored in the configmap| None
 `metadataTries` | The number of times to try requesting metadata. If you would like 2 retries, set metadata-tries to 3. | `3`
 `cordonOnly` | If true, nodes will be cordoned but not drained when an interruption event occurs. | `false`
-`taintNode` | If true, nodes will be tainted when an interruption event occurs. Currently used taint keys are `aws-node-termination-handler/scheduled-maintenance`, `aws-node-termination-handler/spot-itn`, and `aws-node-termination-handler/asg-lifecycle-termination` | `false`
+`drainOnRebalance` | If true, nodes will be drained when a rebalance recommendation notice is received. | `false`
+`taintNode` | If true, nodes will be tainted when an interruption event occurs. Currently used taint keys are `aws-node-termination-handler/scheduled-maintenance`, `aws-node-termination-handler/spot-itn`, `aws-node-termination-handler/asg-lifecycle-termination` and `aws-node-termination-handler/rebalance-recommendation`| `false`
 `jsonLogging` | If true, use JSON-formatted logs instead of human readable logs. | `false`
 `logLevel` | Sets the log level (INFO, DEBUG, or ERROR) | `INFO`
 `enablePrometheusServer` | If true, start an http server exposing `/metrics` endpoint for prometheus. | `false`
--- a/config/helm/aws-node-termination-handler/README.md
+++ b/config/helm/aws-node-termination-handler/README.md
@@ -69,7 +69,6 @@ Parameter | Description | Default
 `webhookTemplateConfigMapKey` | Name of the template file stored in the configmap| None
 `metadataTries` | The number of times to try requesting metadata. If you would like 2 retries, set metadata-tries to 3. | `3`
 `cordonOnly` | If true, nodes will be cordoned but not drained when an interruption event occurs. | `false`
-`drainOnRebalance` | If true, nodes will be drained when a rebalance recommendation notice is received. | `false`
 `taintNode` | If true, nodes will be tainted when an interruption event occurs. Currently used taint keys are `aws-node-termination-handler/scheduled-maintenance`, `aws-node-termination-handler/spot-itn`, `aws-node-termination-handler/asg-lifecycle-termination` and `aws-node-termination-handler/rebalance-recommendation`| `false`
 `jsonLogging` | If true, use JSON-formatted logs instead of human readable logs. | `false`
 `logLevel` | Sets the log level (INFO, DEBUG, or ERROR) | `INFO`
@@ -101,7 +100,8 @@ Parameter | Description | Default
 --- | --- | ---
 `enableScheduledEventDraining` | [EXPERIMENTAL] If true, drain nodes before the maintenance window starts for an EC2 instance scheduled event | `false`
 `enableSpotInterruptionDraining` | If true, drain nodes when the spot interruption termination notice is received | `true`
-`enableRebalanceMonitoring` | If true, cordon nodes when rebalance recommendation is received | `false`
+`enableRebalanceMonitoring` | If true, cordon nodes when the rebalance recommendation notice is received | `false`
+`enableRebalanceDraining` | If true, drain nodes when the rebalance recommendation notice is received | `false`
 `useHostNetwork` | If `true`, enables `hostNetwork` for the Linux DaemonSet. NOTE: setting this to `false` may cause issues accessing IMDSv2 if your account is not configured with an IP hop count of 2 | `true`
 
 ### Kubernetes Configuration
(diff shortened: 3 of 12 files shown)


dennisme:
🎉

config/helm/aws-node-termination-handler/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: aws-node-termination-handler
 description: A Helm chart for the AWS Node Termination Handler
-version: 0.14.2
+version: 0.15.1
```
brycahta, line 4: how did you arrive at this version number?

config/helm/aws-node-termination-handler/README.md
```diff
@@ -69,7 +69,8 @@ Parameter | Description | Default
 `webhookTemplateConfigMapKey` | Name of the template file stored in the configmap| None
 `metadataTries` | The number of times to try requesting metadata. If you would like 2 retries, set metadata-tries to 3. | `3`
 `cordonOnly` | If true, nodes will be cordoned but not drained when an interruption event occurs. | `false`
-`taintNode` | If true, nodes will be tainted when an interruption event occurs. Currently used taint keys are `aws-node-termination-handler/scheduled-maintenance`, `aws-node-termination-handler/spot-itn`, and `aws-node-termination-handler/asg-lifecycle-termination` | `false`
+`drainOnRebalance` | If true, nodes will be drained when a rebalance recommendation notice is received. | `false`
```
brycahta, line 72: For consistency, could you re-name to something like `enableRebalanceDraining` and add this config to **IMDS Mode Configuration** section? Around line 103 in this file

test/e2e/rebalance-recommendation-drain-test
```diff
@@ -0,0 +1,166 @@
...
+  "$CLUSTER_NAME-anth"
+  "$SCRIPTPATH/../../config/helm/aws-node-termination-handler/"
+  --wait
+  --force
+  --namespace kube-system
+  --set instanceMetadataURL="${INSTANCE_METADATA_URL:-"http://$AEMM_URL:$IMDS_PORT"}"
+  --set image.repository="$NODE_TERMINATION_HANDLER_DOCKER_REPO"
+  --set image.tag="$NODE_TERMINATION_HANDLER_DOCKER_TAG"
+  --set enableScheduledEventDraining="false"
+  --set enableSpotInterruptionDraining="true"
```
brycahta, line 40: could you set this to false so we can ensure the drain is a result of a Rebalance msg?

test/e2e/rebalance-recommendation-drain-test
```diff
@@ -0,0 +1,166 @@
...
+elif [[ $tainted -eq 0 ]]; then
+  echo "❌ Worker node was not tainted"
+  fail_and_exit 3
+elif [[ $evicted -eq 0 ]]; then
+  echo "❌ regular-pod-test not was evicted"
+  fail_and_exit 3
+fi
+
+echo "🥑 Waiting for Spot ITN..."
+# Spot ITN will eventually become available since aemm.MockDelaySec will not exceed TAINT_CHECK_CYCLES x TAINT_CHECK_SLEEP
```
brycahta, line 152: nit: should say Rebalance Recommendation

test/e2e/rebalance-recommendation-drain-test
```diff
@@ -0,0 +1,166 @@
...
+for i in $(seq 1 $TAINT_CHECK_CYCLES); do
+    if [[ $(kubectl get deployments regular-pod-test -o=jsonpath='{.status.unavailableReplicas}') -eq 1 ]]; then
+        echo "✅ Verified the regular-pod-test pod was evicted!"
+        echo "✅ Rebalance Recommendation Spot ITN Test Passed $CLUSTER_NAME! ✅"
```
brycahta, line 156: nit: update test name **Rebalance Recommendation Drain Test** for consistency

test/e2e/rebalance-recommendation-drain-test
```diff
@@ -0,0 +1,166 @@
...
+        exit 0
+    fi
+    echo "Assertion Loop $i/$TAINT_CHECK_CYCLES, sleeping for $TAINT_CHECK_SLEEP seconds"
+    sleep $TAINT_CHECK_SLEEP
+done
+
+echo "❌ regular-pod-test was NOT evicted"
+
+echo "❌ Rebalance Recommendation Test Failed $CLUSTER_NAME ❌"
```
brycahta, line 165: nit: update test name **Rebalance Recommendation Drain Test** for consistency

test/e2e/rebalance-recommendation-dry-run-test
```diff
@@ -0,0 +1,139 @@
...
+#   $KUBECONFIG
+#   $NODE_TERMINATION_HANDLER_DOCKER_REPO
+#   $NODE_TERMINATION_HANDLER_DOCKER_TAG
+#   $WEBHOOK_DOCKER_REPO
+#   $WEBHOOK_DOCKER_TAG
+#   $AEMM_URL
+#   $AEMM_VERSION
+
+function fail_and_exit {
+    echo "❌ Rebalance Recommendation Drain test failed $CLUSTER_NAME ❌"
```
brycahta, line 16: nit: Rebalance Recommendation **Dry Run** test

test/e2e/rebalance-recommendation-dry-run-test
```diff
@@ -0,0 +1,139 @@
...
+test_node="${TEST_NODE:-$CLUSTER_NAME-worker}"
+for i in $(seq 1 $TAINT_CHECK_CYCLES); do
+  if [[ $logs -eq 0 && ! -z $(kubectl logs $pod_id -n kube-system | grep -i -e 'would have been cordoned and drained') ]]; then
+      echo "✅ Verified the dryrun logs were executed"
+      logs=1
+  fi
+
+  if [[ $logs -eq 1 ]] && kubectl get nodes $test_node --no-headers | grep -v SchedulingDisabled >/dev/null; then
+      echo "✅ Verified the worker node was not cordoned!"
+      echo "✅ Spot Interruption Dry Run Test Passed $CLUSTER_NAME! ✅"
```
brycahta, line 127: nit: Rebalance Recommendation **Dry Run** test

test/e2e/rebalance-recommendation-dry-run-test
```diff
@@ -0,0 +1,139 @@
...
+  "$SCRIPTPATH/../../config/helm/aws-node-termination-handler/"
+  --wait
+  --force
+  --namespace kube-system
+  --set instanceMetadataURL="${INSTANCE_METADATA_URL:-"http://$AEMM_URL:$IMDS_PORT"}"
+  --set image.repository="$NODE_TERMINATION_HANDLER_DOCKER_REPO"
+  --set image.tag="$NODE_TERMINATION_HANDLER_DOCKER_TAG"
+  --set dryRun="true"
+  --set enableScheduledEventDraining="false"
+  --set enableSpotInterruptionDraining="true"
```
brycahta, line 41: same as above, set this to `false`

cmd/node-termination-handler.go
```diff
@@ -291,7 +291,7 @@ func drainOrCordonIfNecessary(interruptionEventStore *interruptioneventstore.Sto
 		metrics.NodeActionsInc("pre-drain", nodeName, err)
 	}
 
-	if nthConfig.CordonOnly || drainEvent.IsRebalanceRecommendation() {
+	if nthConfig.CordonOnly || drainEvent.IsRebalanceRecommendation() && !nthConfig.DrainOnRebalance {
```
brycahta, line 294: Were you able to test this with `QueueProcessor` mode? The events to listen for are driven by EventBridge rules, so the basic setup for `QueueProcessor` would be:

```
helm upgrade --install aws-node-termination-handler \
  --namespace kube-system \
  --set enableSqsTerminationDraining=true \
  --set queueURL=https://sqs.us-east-1.amazonaws.com/0123456789/my-term-queue \
```

However, this implementation looks like users would need to register EventBridge Rule for RebalanceRecommendation AND apply the config in the helm install command, which is kinda awkward relative to other events:

```
helm upgrade --install aws-node-termination-handler \
  --namespace kube-system \
  --set enableSqsTerminationDraining=true \
  --set drainOnRebalance=true \
  --set queueURL=https://sqs.us-east-1.amazonaws.com/0123456789/my-term-queue \
```

Could you confirm this is the behavior? I'll be trying to think of a cleaner way in the meantime.

brycahta (commented):
Thanks a lot for the work and adding tests! Looks good on a first pass, had some minor comments

cmd/node-termination-handler.go
bwagner5: Can you add parens: 

```
if nthConfig.CordonOnly || (drainEvent.IsRebalanceRecommendation() && !nthConfig.DrainOnRebalance) {
```
I realize it's the same when executed, but it's clearer with the parens.

config/helm/aws-node-termination-handler/README.md
trutx: Oh, sure. I followed the `cordonOnly` implementation but `enableRebalanceDraining` is way more consistent with what we already have.

test/e2e/rebalance-recommendation-drain-test
trutx: Good point. Will do.

test/e2e/rebalance-recommendation-drain-test
trutx: Went too far with the copypasting.

config/helm/aws-node-termination-handler/Chart.yaml
trutx: I manually ran `make ekscharts-sync` and saw that the automatically created PR to `aws/charts` failed because the chart version hadn't been bumped, so I went ahead and bumped it. Not sure if this happens automatically in some build process step.

Since this is not a bug fix PR I bumped the minor version, however now I realise I should've used `0.15.0`. In any case IDK how this decision is made so happy to set the value you see fit.

brycahta: Yeahhh it's not very clear how we bump this--both in documentation and in the commit history, but appreciate that you tried 😅

It'll be easier if our team owns incrementing these chart versions, so **no need for you to make changes to this file in this pr**

#### How we bump versions
If you are interested in the details, we run `make release-prep-<patch,minor,major>` target and the script it calls will bump app and chart versions based on the release type. The target you ran `make ekscharts-sync` is used during the release process to auto-cut a PR to [eks-charts](https://github.com/aws/eks-charts) repo to keep the versions in sync. [Example PR](https://github.com/aws/eks-charts/pull/484)

test/e2e/rebalance-recommendation-drain-test
```diff
@@ -0,0 +1,166 @@
...
+  "$SCRIPTPATH/../../config/helm/aws-node-termination-handler/"
+  --wait
+  --force
+  --namespace kube-system
+  --set instanceMetadataURL="${INSTANCE_METADATA_URL:-"http://$AEMM_URL:$IMDS_PORT"}"
+  --set image.repository="$NODE_TERMINATION_HANDLER_DOCKER_REPO"
+  --set image.tag="$NODE_TERMINATION_HANDLER_DOCKER_TAG"
+  --set enableScheduledEventDraining="false"
+  --set enableSpotInterruptionDraining="false"
+  --set enableRebalanceMonitoring="true"
```
brycahta, line 41: No code change needed, just want to confirm: it's not required to set both `enableRebalanceMonitoring` and `enableRebalanceDraining` to drain on Rebalance events, right? Only the latter?

brycahta (changes requested):
Just remove version bump in `Chart.yaml`-- everything else looks good 👍 

Were you able to confirm the sqs behavior from my previous comment: 
  * Is it necessary to set both `enableSqsTerminationDraining` and `enableRebalanceDraining` to drain on Rebalance in Queue Processor mode?

test/e2e/rebalance-recommendation-drain-test
trutx: No, both are needed. You need the former to enable the monitor [here](https://github.com/aws/aws-node-termination-handler/blob/main/cmd/node-termination-handler.go#L148) and the latter to drain.

IMO it'd be better to follow the other monitors' approach, that is:
* Use `enableRebalanceDraining` param instead of `enableRebalanceMonitoring`
* Drain by default if the monitor is enabled, unless `cordonOnly == true`

This would be consistent with the existing code, but would also mean introducing a breaking change (new parameter, old gets deleted) and also a change in the default behaviour, so I opted for using the existing parameter to enable the monitor and adding a new parameter for enabling the drain.

Again, this and other things call for a bigger refactor that I think should happen some day. That would be the moment to introduce breaking changes, config changes and a major version bump IMHO.

trutx: Also, haven't been able to test it using SQS. The actual code change is in the `drainOrCordonIfNecessary()` func, where an interruption event is passed in. I'm assuming such interruption event would get there just like it does now, regardless of what the source of the event (IMDB or SQS) is.

brycahta: 1. Got it. I agree the current design and param names are a bit awkward as-is-- our team is working on a refactor to help address these issues. One way to bridge the gap now without introducing breaking changes would be to update the enable-the-monitor code you linked to `nthConfig.EnableRebalanceMonitoring || nthConfig.EnableRebalanceDraining` as it's safe to assume if you want draining then you want to monitor for that event. Adds a bit more tech debt for now, but I think it improves the experience-- could you make this change?

2. Your assumptions are correct, I'm just trying to validate if `enableRebalanceDraining` needs to be set to true in order to drain on Rebalance in QueueProcessor mode, which I think is true and that's what we want so we don't break the current default behavior. So all good here.

trutx: Right, no need to check if the event is a rebalance recommendation. Just checking whether drain on rebalance is enabled is enough. Updating it.

trutx: We could add something like
```
	if config.EnableRebalanceDraining {
		config.EnableRebalanceMonitoring = true
	}
```
in `pkg/config/config.go` to spare us from having to set both params. WDYT?

brycahta: I would rather have it more explicit. Instead could you update the enable monitoring logic [here](https://github.com/aws/aws-node-termination-handler/blob/main/cmd/node-termination-handler.go#L148) to be 
`nthConfig.EnableRebalanceMonitoring || nthConfig.EnableRebalanceDraining` ? It should result in the same behavior (not needing to set both params).

trutx: Right, https://github.com/aws/aws-node-termination-handler/pull/400/commits/e99d19aa1bf53a495d1720af4b0ae25bfb9e297f should do it

brycahta (approved):

Issue #390, opened 2021-03-18
Queue mode: missing AWS_REGION attempts to lookup sqs.queue.amazonaws.com and fails


james-callahan:
If I don't set the `AWS_REGION` environment variable, then the server fails to start with:
```
RequestError: send request failed\ncaused by: Post \"https://sqs.queue.amazonaws.com/\": dial tcp: lookup sqs.queue.amazonaws.com on 172.20.0.10:53: no such host
```

This is despite the logs on startup showing me the correct `queue-url` (partially redacted):
```
queue-url: https://us-west-2.queue.amazonaws.com/MY_ACCOUNT_ID/node-termination, 
```

I've worked around this for now by setting `AWS_PROFILE=us-west-2`, but I shouldn't have to.

bwagner5:
Agreed, we will look into this and get to the bottom of it! Thanks for submitting the issue :)

james-callahan:
@jillmon I dont' this should be of type question, it's more of a bug; or if intentional (unlikely), it's an enhancement request.

bwagner5:
I've submitted a PR to fix the region detection within an SQS queue url if no region was specified.  I think it should still be a best practice to set `AWS_REGION` just to be a little more explicit about it. I wouldn't expect SQS queue-urls to change much, but they could.

bwagner5:
released

PR #403, opened 2021-04-09
fixed and improved region detection within queue url

bwagner5:
Issue #, if available:
https://github.com/aws/aws-node-termination-handler/issues/390

Description of changes:
 - Fix issue with SQS queue-url region detection

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -191,6 +191,7 @@ func ParseCliArgs() (config Config, err error) {
 			sess.Config.Region = &config.AWSRegion
 		} else if *sess.Config.Region == "" && config.QueueURL != "" {
 			config.AWSRegion = getRegionFromQueueURL(config.QueueURL)
+			log.Debug().Str("Retrieved AWS region from queue-url: \"%s\"", config.AWSRegion)
 			sess.Config.Region = &config.AWSRegion
 		} else {
 			config.AWSRegion = *sess.Config.Region
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


pkg/config/config.go
```diff
@@ -390,3 +391,14 @@ func isConfigProvided(cliArgName string, envVarName string) bool {
...
+
+func getRegionFromQueueURL(queueURL string) string {
+	for _, partition := range endpoints.DefaultPartitions() {
+		for regionID := range partition.Regions() {
+			if strings.Contains(queueURL, regionID) {
+				return regionID
+			}
+		}
+	}
+	return ""
```
haugenj, line 403: will this lead to an error?

haugenj: if they do end up changing the queue urls or something, maybe we should at least log here

bwagner5: added a debug log

haugenj (approved):
👍

pkg/config/config.go
bwagner5: oh and to answer the first question, yes it will lead to an aws sdk error that the region is undefined

pkg/config/config.go
```diff
@@ -390,3 +392,14 @@ func isConfigProvided(cliArgName string, envVarName string) bool {
 	})
 	return cliArgProvided
 }
+
+func getRegionFromQueueURL(queueURL string) string {
+	for _, partition := range endpoints.DefaultPartitions() {
+		for regionID := range partition.Regions() {
+			if strings.Contains(queueURL, regionID) {
```
james-callahan, line 399: it's not likely; but if someone named their queue `us-west-2` when it was actually in e.g. `ap-southeast-1` then this could have a false positive.

PR #409, opened 2021-04-14
fix error handling for missing instances

haugenj:
Issue #367, if available:

Description of changes:

When an instance is missing the describe-instances call [returns an error](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.DescribeInstances) instead of an empty array. 
> If you specify an instance ID that is not valid, an error is returned. If you specify an instance that you do not own, it is not included in the output.

Now we'll check the error and handle that case 

### DIFF
--- a/pkg/monitor/sqsevent/sqs-monitor.go
+++ b/pkg/monitor/sqsevent/sqs-monitor.go
@@ -17,9 +17,9 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-
 	"github.com/aws/aws-node-termination-handler/pkg/monitor"
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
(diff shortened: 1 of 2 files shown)


pkg/monitor/sqsevent/sqs-monitor.go
```diff
@@ -189,6 +190,11 @@ func (m SQSMonitor) retrieveNodeName(instanceID string) (string, error) {
 		},
 	})
 	if err != nil {
+		// the ec2 client doesn't use aws error codes
```
bwagner5, line 193: this should work:

```
if aerr, ok := err.(awserr.Error); ok ; aerr.Code() == "InvalidInstanceID.NotFound" {
   ...
}
```

Generally InvalidInstanceID.NotFound means the instance just started and isn't describe-able yet. It might be better to do a couple retries rather than just erroring immediately.

brycahta:
Did you try to do a release recently? Looks like ec2-bot is the latest user.

If you know the specific script/target lmk and I'll fix

bwagner5 (approved):
Lgtm

brycahta (approved):
👍🏼

PR #421, opened 2021-04-28
Correct instantiation of AWS session object

### LINKED ISSUE #420: Unable to use STS regional endpoint
I'm trying to spin the NTH up using queue mode in an EKS cluster that doesn't have direct internet access but does have VPC endpoints created for (hopefully) every required service.

This isn't the first controller that I've spun up; every previous one I'm using requires patching in the environment variable `AWS_STS_REGIONAL_ENDPOINTS=regional` in order to make them use `sts.<region>.amazonaws.com` instead of the global `sts.amazonaws.com` endpoint. The former endpoint then resolves to the VPC endpoint I've created and authentication is successful.

I'm getting the following logged when running the NTH:
```
2021/04/26 17:09:25 WRN There was a problem monitoring for events error="WebIdentityErr: failed to retrieve credentials\ncaused by: RequestError: send request failed\ncaused by: Post \"https://sts.amazonaws.com/\": dial tcp 52.94.241.129:443: i/o timeout" event_type=SQS_TERMINATE
2021/04/26 17:11:25 WRN There was a problem monitoring for events error="WebIdentityErr: failed to retrieve credentials\ncaused by: RequestError: send request failed\ncaused by: Post \"https://sts.amazonaws.com/\": dial tcp 52.46.134.192:443: i/o timeout" event_type=SQS_TERMINATE
```
I've set the above environment variable, and also tried removing the `AWS_ENDPOINT` and `AWS_REGION` variables just in case they were confusing things but the Deployment/Pod doesn't appear to be respecting it.

bodgit:
The region needs to be set on the config object before creating the
session so that things like setting `AWS_STS_REGIONAL_ENDPOINTS` can
take effect.

Also, if credentials cannot be acquired, die at that point instead of
carrying on.

Fixes #420

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -33,6 +33,9 @@ import (
 	"github.com/aws/aws-node-termination-handler/pkg/node"
 	"github.com/aws/aws-node-termination-handler/pkg/observability"
 	"github.com/aws/aws-node-termination-handler/pkg/webhook"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/sqs"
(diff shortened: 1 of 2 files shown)


haugenj (approved):
👍 LGTM, thanks!

bwagner5:
I'm good with your changes, but rather than making users specify: `export AWS_STS_REGIONAL_ENDPOINTS regional`  we can just hard code the config to use regional endpoints. I don't think there's a reason to use the global endpoint w/ NTH.  If there's something I'm missing with that assumption, then a second option would be to add the `AWS_STS_REGIONAL_ENDPOINTS` var in the helm chart's values.yaml and default it to `regional`. 

```
cfg := aws.NewConfig().
			WithRegion(nthConfig.AWSRegion).
			WithEndpoint(nthConfig.AWSEndpoint).
			WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint)
		sess := session.Must(session.NewSessionWithOptions(session.Options{
			Config:            *cfg,
			SharedConfigState: session.SharedConfigEnable,
		}))
```

bodgit:
It was purely a case of not changing existing behaviour. You can do exactly as you describe and force the regional endpoint when creating the session object; I've seen some controllers/drivers do that. If you're happy to have that slight change in behaviour then I can tweak the PR. That would be one less Kustomization patch I have to have :wink:

bwagner5:
Cool, I’m good with hard coding it in the session! Mind adding to this pr?

bodgit:
Done

bwagner5 (approved):
nice! Thank you! LGTM

Issue #412, opened 2021-04-15
Pod crashing when podTerminationGracePeriod is reached


heschlie:
I've noticed that when the podTerminationGracePeriod is reached the NTH looks to crash, I'm unsure if it is my settings or an actual problem with NTH. We unfortunately have some pods that sometimes seem to hang around indefinitely, but I had figured after reaching the grace period it would simply forcibly kill those pods.

```
aws-node-termination-handler arguments:
dry-run: false,
node-name: <snip>,
metadata-url: http://169.254.169.254,
kubernetes-service-host: 172.20.0.1,
kubernetes-service-port: 443,
delete-local-data: true,
ignore-daemon-sets: true,
pod-termination-grace-period: 120,
node-termination-grace-period: 120,
enable-scheduled-event-draining: false,
enable-spot-interruption-draining: false,
enable-sqs-termination-draining: true,
enable-rebalance-monitoring: false,
metadata-tries: 3,
cordon-only: false,
taint-node: false,
json-logging: false,
log-level: info,
webhook-proxy: ,
webhook-headers: <not-displayed>,
webhook-url: ,
webhook-template: <not-displayed>,
uptime-from-file: ,
enable-prometheus-server: true,
prometheus-server-port: 9092,
aws-region: us-east-1,
queue-url: <snip>,
check-asg-tag-before-draining: true,
managed-asg-tag: aws-node-termination-handler/managed,
aws-endpoint: ,
```

I unfortunately don't have the logs, but the last bit of output is always the list of pods it is evicting, followed by the error `There was a problem while trying to cordon and drain the node` saying it reached the pod grace period. It then exit(1) and a new pod is spun up, which sees the old message, and picks up where the last one left off. It seems like when it is trying to forcibly kill the nuisance pods (assuming it actually does this) is when the NTH dies, as then the next one comes in all of the pods are gone and it simply wraps up the lifecycle event and deletes the SQS message.

heschlie:
I looks like this is happening here https://github.com/aws/aws-node-termination-handler/blob/main/cmd/node-termination-handler.go#L318

I'm guessing when the grace period is reached in the `drain.RunNodeDrain(n.drainHelper, node.Name)` from the k8s pkg is reached it forcibly deletes the pods but returns an error. If that's the case it might be possible to check that error and not have it exit(1)?

haugenj:
I think the os.exit(1) was written for imds mode, and probably shouldn't be executed when we're running in queue-processor mode

[...]

PR #422, opened 2021-04-30
don't exit(1) in queue processor mode

heschlie:
Issue #, if available: #412

Description of changes: Don't `exit(1)` on drain errors when in queue processor mode

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -302,7 +302,7 @@ func drainOrCordonIfNecessary(interruptionEventStore *interruptioneventstore.Sto
 	if nthConfig.CordonOnly || (drainEvent.IsRebalanceRecommendation() && !nthConfig.EnableRebalanceDraining) {
 		cordonNode(node, nodeName, drainEvent, metrics, recorder)
 	} else {
-		cordonAndDrainNode(node, nodeName, metrics, recorder)
+		cordonAndDrainNode(node, nodeName, metrics, recorder, nthConfig.EnableSQSTerminationDraining)
 	}
 
 	interruptionEventStore.MarkAllAsProcessed(nodeName)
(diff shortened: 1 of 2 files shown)


heschlie:
So, I'm having trouble with the `run-report-card-test.sh` test. It seems like there might be a couple errors in it, but I'm wondering how it ever worked and if I'm just dumb:

shouldn't [this line](https://github.com/aws/aws-node-termination-handler/blob/main/test/go-report-card-test/run-report-card-test.sh#L15) be...
```bash
if [[ ! $(docker run -it -v $SCRIPTPATH/../..:/app go-report-card-cli /go/bin/goimports -l /app) ]]; then
```

As `goimports` should only return an error if it finds offending imports?

Also, [this line](https://github.com/aws/aws-node-termination-handler/blob/main/test/go-report-card-test/run-report-card-test.sh#L22) should probably exclude the `vendor` directory? Also it seems that `goreportcard-cli` needs the directory specified in order to skip the vendor dir, this wasn't erroring locally but it was noisy.

I've made these changes, please let me know if I should revert them. Again I feel like if it was a problem it would have shown up before and I'm just missing something.

brycahta:
Hey @heschlie , thanks for the PR!

We have an open issue #424 to fix go-report-card as we saw it start to break recently, so no worries if it's failing on your changes. I should get the fix completed early next week

heschlie:
Hey @brycahta thanks for the heads up, as for the E2E tests it looks to be failing on the first E2E script:

```
======================================================================================================
🥑 Running assertion script asg-lifecycle-sqs-test
======================================================================================================
Starting ASG Lifecycle SQS Test for Node Termination Handler
+ helm upgrade --install nth-test-2377abf5-localstack /home/runner/work/aws-node-termination-handler/aws-node-termination-handler/test/e2e/../../config/helm/localstack/ --wait --namespace default --set 'nodeSelector.kubernetes\.io/hostname=nth-test-2377abf5-control-plane' --set defaultRegion=us-east-1
Release "nth-test-2377abf5-localstack" does not exist. Installing it now.
Error: timed out waiting for the condition
```

Is there a way for me to get more detail on why helm (I'm assuming) timed out?

Issue #424, opened 2021-05-03
Fix go-report-card issues


brycahta:
`go-report-card` is failing right now. 

Will switch to using `golangci-lint` like [karpenter](https://github.com/awslabs/karpenter/blob/main/Makefile#L37)

PR #427, opened 2021-05-10
Upgrade to Go 1.16 and fix report-card

brycahta:
Issue #, if available: #424

Description of changes:
* Upgrade to go 1.16
* Fixed files failing go-report-card

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: NTH Continuous Integration and Release
 on: [push, pull_request, workflow_dispatch]
 
 env:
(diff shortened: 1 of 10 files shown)


brycahta:
Same issue that was brought up in a past PR. 

`ineffassign` fails with 0% , but running locally it returns 100%. If I run **just ineffassign locally** it returns no results meaning no issue. My guess is re-running the workflow may pass.

I think the deprecated [gometalinter](https://github.com/alecthomas/gometalinter) which is being used to run `ineffassign` [here](https://github.com/gojp/goreportcard/blob/master/check/ineffassign.go#L21) is a potential root cause

PR #432, opened 2021-05-13
always do license check and use dockerhub for go-report-card image pulls

bwagner5:
Issue #, if available:
N/A

Description of changes:
 - ECR rate limits pulls for unauthenticated users, so pull go-report-card image from dockerhub 
 - Always run license tests

### DIFF
--- a/test/go-report-card-test/Dockerfile
+++ b/test/go-report-card-test/Dockerfile
@@ -1,4 +1,4 @@
-FROM public.ecr.aws/bitnami/golang:latest
+FROM golang:latest
 
 WORKDIR /app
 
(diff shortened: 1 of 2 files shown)


test/go-report-card-test/Dockerfile
```diff
@@ -12,4 +12,4 @@ RUN go get github.com/gojp/goreportcard/cmd/goreportcard-cli
 
 RUN go get -u golang.org/x/tools/cmd/goimports
 
-CMD $GOPATH/bin/goreportcard-cli -v -t 90
\ No newline at end of file
+CMD $GOPATH/bin/goreportcard-cli -v -t 90
```
brycahta, line 15: does this threshold not mean anything/can it be removed? It's set to 93 in the script itself and it seems to adhere to that than this param

bwagner5: it's the default threshold if you don't specify any command when calling docker run.

brycahta (approved):

PR #440, opened 2021-05-24
[ec2metadata] Switch IMDSv2 token warning message to debug

hwoarang:
If IMDSv2 is unreachable or disabled on the EC2 instances, the logs
are filled with such messages without providing any real value or action
as the application fallbacks to IMDSv1 anyway.  We can move such messages
to the debug level in order to try and keep the log messages clean for
any warnings that might be missed with this flooding.

These messages probably fail into the debug category anyway as they are
mostly helpful when debugging the application.


bwagner5 (approved):
Agreed, lgtm! Thanks!

PR #447, opened 2021-06-02
add pods to webhook event and log pods on node

bwagner5:
Issue #, if available:
N/A

Description of changes:
 - Add pod list to drain event (before it was only being added if cordon-only was enabled or rebalance events in IMDS mode)

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -309,18 +309,21 @@ func drainOrCordonIfNecessary(interruptionEventStore *interruptioneventstore.Sto
 	}
 
 	if nthConfig.CordonOnly || (drainEvent.IsRebalanceRecommendation() && !nthConfig.EnableRebalanceDraining) {
-		cordonNode(node, nodeName, drainEvent, metrics, recorder)
+		err = cordonNode(node, nodeName, drainEvent, metrics, recorder)
 	} else {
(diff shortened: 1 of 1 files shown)


cmd/node-termination-handler.go
```diff
@@ -313,6 +313,15 @@ func drainOrCordonIfNecessary(interruptionEventStore *interruptioneventstore.Sto
 	} else {
 		cordonAndDrainNode(node, nodeName, metrics, recorder, nthConfig.EnableSQSTerminationDraining)
 	}
+	podNameList, err := node.FetchPodNameList(nodeName)
```
haugenj, line 316: Looks like this will slightly change the behavior - previously if the node didn't exist then this code wouldn't execute, but now that you've moved this block out of the `else` block from line 349, this can get ran when the node is gone.

Perhaps we should return errors from the `cordonNode` and `cordonAndDrainNode` funcs and then check against them here?

imuqtadir: Do we want to show the list of of pods during drain? My understanding was that these pods are being drained anyway so why  care. I guess maybe they just want to maintain list of pods getting drained, is it?

bwagner5: yeah, I don't really see a reason not to. Only users that want it have to put it into their webhook template.

haugenj (approved):
lgtm

PR #452, opened 2021-06-07
Upgrade deps

bwagner5:
Issue #, if available:
n/a

Description of changes:
 - Upgrade dependencies


pkg/monitor/scheduledevent/scheduled-event-monitor_internal_test.go
```diff
@@ -91,28 +79,34 @@ func TestUncordonAfterRebootPreDrainSuccess(t *testing.T) {
...
 
 	client := fake.NewSimpleClientset()
 	//nolint:errcheck
-	client.CoreV1().Nodes().Create(&v1.Node{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: nodeName,
-		},
-		Spec: v1.NodeSpec{
-			Unschedulable: true,
+	_, err := client.CoreV1().Nodes().Create(context.Background(),
```
haugenj, line 95: shouldn't the `//nolint:errcheck` let us ignore `err` here? There's times where you're catching the `err` value and other times you ignore it, I think we should make it consistent one way or the other

bwagner5: yeah I think I prefer to just check the errs in all cases since it's the same number of lines, will update.

haugenj (approved):
/lgtm

PR #453, opened 2021-06-07
Upgrade to golangci-lint

brycahta:
Issue #, if available: #424

Description of changes:
* removed go-report-card test, targets, and GH Action & replaced with golangci-lint
* fixed issues after running `golangci-lint run` locally
* majority of call-outs were `Error return value of rw.Write is not checked`
  * used my judgement here to determine when to ignore this (lots of test files)
  * feel free to comment/request changes if any areas should be checked

* not sure if I need to add to `go.mod` since we will use golangci-lint GHA and not a test folder/file like go-report-card


Makefile
```diff
@@ -153,7 +153,7 @@ release: build-binaries build-docker-images push-docker-images generate-k8s-yaml
 
 release-windows: build-binaries-windows build-docker-images-windows push-docker-images-windows upload-resources-to-github-windows
 
-test: spellcheck shellcheck unit-test e2e-test compatibility-test license-test go-report-card-test helm-sync-test helm-version-sync-test helm-lint
+test: spellcheck shellcheck unit-test e2e-test compatibility-test license-test helm-sync-test helm-version-sync-test helm-lint
```
bwagner5, line 156: add go-linter here?

README.md
```diff
@@ -12,9 +12,6 @@
   <a href="https://opensource.org/licenses/Apache-2.0">
     <img src="https://img.shields.io/badge/License-Apache%202.0-ff69b4.svg" alt="license">
   </a>
-  <a href="https://goreportcard.com/report/github.com/aws/aws-node-termination-handler">
```
bwagner5, line 15: :( A+

bwagner5 (commented):
lgtm, just the one comment on adding linter to the test make target. 

You can leave out of go.mod.  karpenter takes the approach of using a second go.mod for tools like this: https://github.com/awslabs/karpenter/tree/main/tools 
so maybe we could do something like that eventually, but it can wait.

Makefile
brycahta: yeah, I can. Was debating if we should put it before the `go.mod` addition. It works on my machine though so i'll send!

README.md
brycahta: haha end of an era

bwagner5 (approved):
lgtm!

PR #456, opened 2021-06-09
helm: Use hostPort only with hostNetwork

hwoarang:
There is no need to use the hostPort option if we are not directly
connected to the hostNetwork. Using the hostPort option makes that port
unavailable for binding to any other service on the node which is
not really desirable if the pod itself is not on the host network.

ps: This is a copy of https://github.com/aws/eks-charts/pull/527


bwagner5 (approved):
LGTM! Thanks!

PR #464, opened 2021-07-05
fixes #423: Populate pod list in webhook notification

### LINKED ISSUE #423: Unable to get pod names in webhook notification.
As per the PR https://github.com/aws/aws-node-termination-handler/pull/388, we should be able to get the Pod names by including the following in the webhook template: `Pods: {{ .Pods }}`. However, it returns an empty array `Pods: []`.
Slack notification:
```
EventID: spot-itn-xxxx
Kind: SPOT_ITN
Description: Spot ITN received. Instance will be interrupted at 2021-05-03T08:30:49Z
Start Time: 2021-05-03 08:30:49 +0000 UTC
Node: ip-10-xx-xx-xx.eu-west-1.compute.internal
Pods: []
```

This is tested on tag: `v1.13.0`
Helm values used:
```
affinity:
  nodeAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      nodeSelectorTerms:
      - matchExpressions:
        - key: kubernetes.io/os
          operator: In
          values:
          - linux
        - key: kubernetes.io/arch
          operator: In
          values:
          - amd64
        - key: eks.amazonaws.com/compute-type
          operator: NotIn
          values:
          - fargate
deleteLocalData: true
enablePrometheusServer: true
enableScheduledEventDraining: true
enableSpotInterruptionDraining: true
ignoreDaemonSets: true
jsonLogging: true
nodeTerminationGracePeriod: 120
podAnnotations:
  co.elastic.logs/enabled: "true"
podMonitor:
  create: true
podTerminationGracePeriod: 60
priorityClassName: system-node-critical
rbac:
  pspEnabled: true
updateStrategy:
  rollingUpdate:
    maxUnavailable: 100%
  type: RollingUpdate
webhookTemplate: |
  {
    "blocks": [
      {
        "type": "section",
        "text": {
          "type": "mrkdwn",
          "text": ":boom: Instance interruption on cluster :boom:"
        }
      },
      {
        "type": "divider"
      },
      {
        "type": "context",
        "elements": [
          {
            "type": "mrkdwn",
            "text": "*EventID:* {{ .EventID }}\n*Kind:* {{ .Kind }}\n*Description:* {{ .Description }}\n*Start Time:* {{ .StartTime }}\n*Node:* {{ .NodeName }}\n*Pods:* {{ .Pods }}"
          }
        ]
      }
    ]
  }
webhookURL: xxx
```

vermaabhay:
Issue #, if available:
#423

Description of changes:
- `cordonNode` & `cordonAndDrainNode` returns nil on successful execution and hence the `podName` and `node.LogPods` is not subsequently called in current implementation.

- Pod list should be fetched before `cordonNode` or `cordonAndDrainNode`. After successful `cordonAndDrainNode` operation the running pods will be evicted from the node and pod list be will empty in that case.

- Even if `cordonNode` or `cordonAndDrainNode` fails one still needs the list of running pods in notification message to capture which all pods will be impacted by the interruption event. Helpful in debugging issues.

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -322,6 +322,11 @@ func drainOrCordonIfNecessary(interruptionEventStore *interruptioneventstore.Sto
         if err != nil {
                 log.Err(err).Msgf("Unable to fetch running pods for node '%s' ", nodeName)
         }
+        drainEvent.Pods = podNameList
+        err = node.LogPods(podNameList, nodeName)
+        if err != nil {
+                log.Err(err).Msg("There was a problem while trying to log all pod names on the node")
+        }
 
 	if nthConfig.CordonOnly || (drainEvent.IsRebalanceRecommendation() && !nthConfig.EnableRebalanceDraining) {
 		err = cordonNode(node, nodeName, drainEvent, metrics, recorder)
@@ -329,20 +334,19 @@ func drainOrCordonIfNecessary(interruptionEventStore *interruptioneventstore.Sto
 		err = cordonAndDrainNode(node, nodeName, metrics, recorder, nthConfig.EnableSQSTerminationDraining)
 	}
         
-        drainEvent.Pods = podNameList
-        err = node.LogPods(podNameList, nodeName)
-        if err != nil {
-                log.Err(err).Msg("There was a problem while trying to log all pod names on the node")
-        }
-
-	interruptionEventStore.MarkAllAsProcessed(nodeName)
 	if nthConfig.WebhookURL != "" {
 		webhook.Post(nodeMetadata, drainEvent, nthConfig)
 	}
-	if drainEvent.PostDrainTask != nil {
-		runPostDrainTask(node, nodeName, drainEvent, metrics, recorder)
-	}
-	<-interruptionEventStore.Workers
+
+        if err != nil {
+                <-interruptionEventStore.Workers
+        } else {
+	        interruptionEventStore.MarkAllAsProcessed(nodeName)
+	        if drainEvent.PostDrainTask != nil {
+		        runPostDrainTask(node, nodeName, drainEvent, metrics, recorder)
+	        }
+	        <-interruptionEventStore.Workers
+        }
 
 }
 
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -318,23 +318,22 @@ func drainOrCordonIfNecessary(interruptionEventStore *interruptioneventstore.Sto
 		runPreDrainTask(node, nodeName, drainEvent, metrics, recorder)
 	}
 
+        podNameList, err := node.FetchPodNameList(nodeName)
+        if err != nil {
+                log.Err(err).Msgf("Unable to fetch running pods for node '%s' ", nodeName)
+        }
+
 	if nthConfig.CordonOnly || (drainEvent.IsRebalanceRecommendation() && !nthConfig.EnableRebalanceDraining) {
 		err = cordonNode(node, nodeName, drainEvent, metrics, recorder)
 	} else {
 		err = cordonAndDrainNode(node, nodeName, metrics, recorder, nthConfig.EnableSQSTerminationDraining)
 	}
-
-	if err != nil {
-		podNameList, err := node.FetchPodNameList(nodeName)
-		drainEvent.Pods = podNameList
-		if err != nil {
-			log.Err(err).Msgf("Unable to fetch running pods for node '%s' ", nodeName)
-		}
-		err = node.LogPods(podNameList, nodeName)
-		if err != nil {
-			log.Err(err).Msg("There was a problem while trying to log all pod names on the node")
-		}
-	}
+        
+        drainEvent.Pods = podNameList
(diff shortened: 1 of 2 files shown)


cmd/node-termination-handler.go
```diff
@@ -318,23 +318,22 @@ func drainOrCordonIfNecessary(interruptionEventStore *interruptioneventstore.Sto
 		runPreDrainTask(node, nodeName, drainEvent, metrics, recorder)
 	}
 
+        podNameList, err := node.FetchPodNameList(nodeName)
```
bwagner5, line 321: I'm okay with your change to fix the pod list (thanks for that btw), but I'm not sure I agree that a webhook should be sent if `cordonNode/cordonAndDrainNode` return an err.  We weren't checking the err before it looks like (whoops), but I think it should check the err right after the `cordon*` if-branches and  if the err is not nil, then signal the interruptionEventStore.Workers  (not executing the PostDrainTask or webhook).  

I'm happy to make that change after this PR is merged since that seems to have been a latent bug.

bwagner5: it looks like go-linter is failing and the Scheduled Maintenance Events dry-run test for IMDS mode e2e test is failing:

```
golangci-lint run --timeout=60m
cmd/node-termination-handler.go:327:3: ineffectual assignment to `err` (ineffassign)
		err = cordonNode(node, nodeName, drainEvent, metrics, recorder)
		^
cmd/node-termination-handler.go:329:3: ineffectual assignment to `err` (ineffassign)
		err = cordonAndDrainNode(node, nodeName, metrics, recorder, nthConfig.EnableSQSTerminationDraining)
		^
```

I think if you do what I mention above, it will fix this one. 

```
goroutine 51 [running]:
k8s.io/client-go/kubernetes/typed/core/v1.(*pods).List(0xc0000cefa0, 0x24fb2e8, 0xc0000ae018, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc00040af30, ...)
	/go/pkg/mod/k8s.io/client-go@v0.21.1/kubernetes/typed/core/v1/pod.go:95 +0xe7
github.com/aws/aws-node-termination-handler/pkg/node.Node.fetchAllPods(0x1, 0xc0000400ca, 0x18, 0xc00004a016, 0x46, 0x101, 0xc0000402b8, 0x9, 0xc0000461f8, 0x3, ...)
	/node-termination-handler/pkg/node/node.go:499 +0x175
github.com/aws/aws-node-termination-handler/pkg/node.Node.FetchPodNameList(0x1, 0xc0000400ca, 0x18, 0xc00004a016, 0x46, 0x101, 0xc0000402b8, 0x9, 0xc0000461f8, 0x3, ...)
	/node-termination-handler/pkg/node/node.go:366 +0x85
main.drainOrCordonIfNecessary(0xc000152540, 0xc000794000, 0x1, 0xc0000400ca, 0x18, 0xc00004a016, 0x46, 0x101, 0xc0000402b8, 0x9, ...)
	/node-termination-handler/cmd/node-termination-handler.go:321 +0x178
created by main.main
	/node-termination-handler/cmd/node-termination-handler.go:238 +0x12c8
```

I'm not entirely sure why this is happening. I'd have to dig into it more.

vermaabhay: IMHO webhook notification is important to notify of an interruption event even if NTH fails to cordon/drain the node for any reason. Additionally, NTH can send the reason for cordon/drain failure in webhook notification itself as another field in coming releases.

vermaabhay: It seems `fetchAllPods` tried to get the list of pods from a lost node. Not sure if I am entirely correct here for the below issue.
`panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x137c7a7]`

bwagner5: I see your point on sending a webhook even if the drain fails. Ideally, that won't happen very often. I would like for the webhook to say that it failed though.

vermaabhay: `drainHelper.Client` was not initialised when dry-run was set to true and therefore it was referencing an invalid memory address. Since `fetchAllPods` requires `drainHelper.Client` to be initialised before calling the core v1 pod api. 
Now, all test cases passed after this change. Without this change all test cases with dry-run set to true were failing.

I will open another feature request for webhook changes to send the status of cordon/drain failures in the notification.

vermaabhay: Incase of `dry-run` as we are not initialising `drainHelper.Client` so `fetchAllPods` should also be not called.
This should fix the unit as well as e2e test cases. I tested this on my local for all e2e cases.

bwagner5: 👍  That looks like the right thing to do! :)   I'll take a closer look at the full code a little later, but I did pre-approve for tests to run. Thanks for doing this work and contributing to NTH!

vermaabhay: Hi @bwagner5,
If the changes look good to you, please ship it with v1.13.2.

bwagner5 (approved):
lgtm!

PR #479, opened 2021-08-17
Update README for instance state changes

boffbowsh:
NTH's SQS processor does automatically support instance state changes for EC2 instances. If a node is Terminated, either by a user or by an AWS scheduled event, we still have a small window in which to begin the drain. 

The documentation as it stands doesn't set up the EventBridge rule to enable this. I've added it to the basic setup guide here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,7 +263,7 @@ $ aws sqs create-queue --queue-name "${SQS_QUEUE_NAME}" --attributes file:///tmp
 
 #### 4. Create Amazon EventBridge Rules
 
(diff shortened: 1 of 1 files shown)


bwagner5 (approved):
lgtm! Thank you!!

gabegorelick:
Apologies if there is a better forum for this question, but I assume this feature only makes sense if https://kubernetes.io/blog/2021/04/21/graceful-node-shutdown-beta/ is disabled? Otherwise, it seems like the kubelet running on the node is in a better position to react to shutdowns.

PR #486, opened 2021-08-24
Release Faster

bwagner5:
Don't run builds and tests on a release since they've been run at least twice before.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,7 +126,6 @@ jobs:
   releaseLinux:
     name: Release Linux
     runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


brycahta (changes requested):
Update `needs:` block to remove *build* jobs. Ex: 

```
  releaseLinux:
    name: Release Linux
    runs-on: ubuntu-20.04
    needs: [buildWindows, buildWindowsDocker, buildLinux, buildLinuxDocker, fastTests]

```

brycahta (approved):
looks fast to me 🚀

brycahta:
But this should be running the tests?

brycahta (approved):
looks good pending tests 👍

PR #489, opened 2021-09-02
fix docker layer caching

bwagner5:
## Issue #, if available:
N/A

## Description of changes:
Reduce build times by leveraging docker caching layers. Previously, go mod downloads were not cached between os/arch variants because the build args changed which resulted in docker invalidating the go mod download layer.


brycahta (approved):
Nice!! `Build Linux Binaries` went from 21m 51s to 9m 35s!  🚀

PR #490, opened 2021-09-10
Add cordon reason in log messages for cordon operations

### LINKED ISSUE #386: logLevel info should state the reason for cordoning
The logs in info mode should state the reason for cordoning/termination.

snay2:
Issue #, if available: Fixes #386

Description of changes:

`drainEvent.Description` indicates the reason for needing to cordon and/or drain a node, as well as some other details like the time the event will occur. Currently, the following reasons are possible:

* [Spot ITN](https://github.com/aws/aws-node-termination-handler/blob/main/pkg/monitor/spotitn/spot-itn-monitor.go#L94)
* [Scheduled event](https://github.com/aws/aws-node-termination-handler/blob/main/pkg/monitor/scheduledevent/scheduled-event-monitor.go#L105)
* [Rebalance recommendation](https://github.com/aws/aws-node-termination-handler/blob/main/pkg/monitor/rebalancerecommendation/rebalance-recommendation-monitor.go#L92)
* [SQS event](https://github.com/aws/aws-node-termination-handler/blob/4a6f45339b20fc22f98a59fec3292684b8798612/pkg/monitor/sqsevent/sqs-monitor.go#L108-L124), which can include ASG terminations, EC2 state change, Spot ITN, or rebalance recommendations

These changes pass the cordon reason down to each level where it may be needed for a log message but do not otherwise change the cordon/drain behavior.

An example log message, before the change:
```
2021/09/13 18:28:28 INF Node successfully cordoned and drained node_name=ip-192-168-0-1.ec2.internal
```

After:

```
2021/09/13 18:28:28 INF Node successfully cordoned and drained node_name=ip-192-168-0-1.ec2.internal reason="ASG Lifecycle Termination event received. Instance will be interrupted at 2021-09-13 18:28:16 +0000 UTC \n"
```

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -362,7 +362,7 @@ func runPreDrainTask(node node.Node, nodeName string, drainEvent *monitor.Interr
 }
 
 func cordonNode(node node.Node, nodeName string, drainEvent *monitor.InterruptionEvent, metrics observability.Metrics, recorder observability.K8sEventRecorder) error {
-	err := node.Cordon(nodeName, drainEvent.Kind)
+	err := node.Cordon(nodeName, drainEvent.Description)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			log.Err(err).Msgf("node '%s' not found in the cluster", nodeName)
@@ -373,7 +373,7 @@ func cordonNode(node node.Node, nodeName string, drainEvent *monitor.Interruptio
 		}
 		return err
 	} else {
-		log.Info().Str("node_name", nodeName).Str("reason", drainEvent.Kind).Msg("Node successfully cordoned")
+		log.Info().Str("node_name", nodeName).Str("reason", drainEvent.Description).Msg("Node successfully cordoned")
 		metrics.NodeActionsInc("cordon", nodeName, err)
 		recorder.Emit(nodeName, observability.Normal, observability.CordonReason, observability.CordonMsg)
 	}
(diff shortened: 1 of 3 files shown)


cmd/node-termination-handler.go
```diff
@@ -373,15 +373,15 @@ func cordonNode(node node.Node, nodeName string, drainEvent *monitor.Interruptio
 		}
 		return err
 	} else {
-		log.Info().Str("node_name", nodeName).Msg("Node successfully cordoned")
+		log.Info().Str("node_name", nodeName).Str("reason", drainEvent.Kind).Msg("Node successfully cordoned")
```
bwagner5, line 376: rather than using `drainEvent.Kind` can we use `drainEvent.Description`? On one hand, `Kind` seems like a minimal amount of information which is pretty nice. But the problem is that `SQS_TERMINATION` doesn't really have any meaning. When NTH is running in `sqs-processor` mode, all the reasons will be `SQS_TERMINATION` even when the SQS_TERMINATION event could actually be a SPOT_ITN, SPOT_RBR, ASG_LIFECYCLE_HOOK, etc.  

wdut?

bwagner5 (changes requested):
looks like one of the unit tests needs updated with the new function signature.

```
pkg/node/node_test.go:64:28: not enough arguments in call to tNode.CordonAndDrain
	have (string)
	want (string, string)
```

cmd/node-termination-handler.go
snay2: I agree, the SQS case specifically needs more detail. `Description` looks like a much better choice. Will update the PR to use that instead.

bwagner5 (approved):
lgtm, thanks!

PR #491, opened 2021-09-11
Add metrics documentation

artificial-aidan:
Description of changes:

Add documentation of the available metrics. Took me longer than I care to admit to figure out which metrics were available because they aren't namespaced with `aws_node_termintation` or similar


docs/metrics.md
```diff
@@ -0,0 +1,10 @@
+# Metrics
+
+## Prometheus Metrics
+
+Available Prometheus metrics:
+
+| Metric name    | Description                           |
+| -------------- | ------------------------------------- |
+| `actions_node` | Number of actions per node            |
+| `events_error` | Number of errors in events processing |
```
snay2, line 10: If I understand correctly, we also emit two other metrics, defined in the [observability package](https://github.com/aws/aws-node-termination-handler/blob/main/pkg/observability/opentelemetry.go#L29-L35). Can you include those as well?

snay2 (changes requested):
Welcome, and thank you for your contribution!

Since this is such a short section, we'd prefer to keep it in a file that gets more traffic, like the main [README.md](https://github.com/aws/aws-node-termination-handler/blob/main/README.md) (perhaps after the "Building" section). That will make it more likely to stay up-to-date in the future.

docs/metrics.md
artificial-aidan: I think those are the metric labels. The [registration function](https://github.com/aws/aws-node-termination-handler/blob/ccaae0544f0971ca80f501734fdb5b70952a3235/pkg/observability/opentelemetry.go#L106-L125) only registers the 2.

snay2 (approved):
lgtm, thank you!

Issue #498, opened 2021-10-12
An event store bug slows event handling over time


mechanical-fish:
The `main()` event loop runs once per second and repeatedly calls `GetActiveEvent()` to find nodes that need to be cordoned or drained. It marks each event InProgress before handling it. There is no code which un-sets InProgress.

GetActiveEvent() returns the first event it finds that isn't marked NodeProcessed. It does not check the InProgress flag.

A run of the event loop stops immediately if GetActiveEvent() returns an InProgress event. 

There are evidently code paths which leave an event marked InProgress without marking it NodeProcessed. This happens slowly but routinely as NTH runs with a reasonably large EKS cluster.

I believe that result is that the event store gradually fills up with InProgress events which, when GetActiveEvent() returns one, immediately halt the event loop until its next iteration one second later. The fact that GetActiveEvent() chooses what to return by iterating over a Go map, which traverses the map in an undefined and semi-random order, helps disguise this problem -- if map iteration were deterministic the event loop would eventually halt forever. But instead, as more and more stale InProgress events accumulate, the odds of getting lucky and handling a new event before it becomes moot grow lower and lower.

I have a PR which has been helpful for us, which I'll link to this issue.

PR #499, opened 2021-10-12
Fix slowdown and stoppage in the main event loop

mechanical-fish:
Issue #, if available: #498

Description of changes:

- Prevent the event loop from receiving stale events marked InProgress and exiting immediately, before reaching urgent events that need handling.

- Periodically log the size of the event queue.

- Periodically "garbage-collect" the event queue to trim already-handled events and slow the rate of its unbounded growth.

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -222,12 +222,14 @@ func main() {
 	log.Info().Msg("Started watching for event cancellations")
 
 	var wg sync.WaitGroup
+
 	for range time.NewTicker(1 * time.Second).C {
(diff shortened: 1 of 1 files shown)


mechanical-fish:
I think you're right that the correct thing to do is to delete the event completely if `drainOrCordonIfNecessary`'s call to `drainOrCordonNode` or `cordonNode` return an error. After all, it looks like because the event has `InProgress` set it will never do anything again, and i'm definitely still seeing residual uncollected stale events after the garbage-collector runs; this might help eliminate that.

PR #506, opened 2021-10-13
Improve API efficiency and handling of unmanaged nodes

mechanical-fish:
Description of changes:

While running an AWS account with a number of EKS clusters and AutoScaling groups, it's important to rapidly discard incoming SQS messages from EventBridge which target unrelated ASGs. The existing code offers the `CheckASGTagBeforeDraining` option -- known as `CheckIfManaged` in the code itself -- which rejects such messages. But it does not handle them; it just leaves them on the SQS queue, where they will usually be fetched several times until the underlying instances are finally terminated and NTH finally deletes them. 

This PR fixes that problem by immediately dropping SQS messages that target instances which do not have the `ManagedAsgTag` set.

While implementing this I noticed that the implementation of `CheckIfManaged` is very inefficient:

- After looking up the instance record with `ec2.DescribeInstances` the code calls `autoscaling.DescribeAutoScalingInstances` to look up the ASG for the instance and fill it into the `InterruptionEvent` record.
- Then to check the tag it makes the very same call to `autoscaling.DescribeAutoScalingInstances` _again_ to look up the ASG, then calls `autoscaling.DescribeTagsPages` to list the tags on the ASG and check if one of those tags matches the configured `ManagedAsgTag`.

I've realized that none of these API calls are really necessary for most, if not all, users:

- I'm not sure it's ever necessary to call `autoscaling.DescribeAutoScalingInstances` now -- if I'm reading the docs right, every ASG instance has a built-in `aws:autoscaling:groupName` tag that is set by AutoScaling itself, and we can just read the value of that tag.
- ASG has supported tag propagation for a while and my company has it switched on for all tags in all ASGs. (_Does anyone not do this?_ 😆 ) That means that the configured `ManagedAsgTag` is also on the instances, and is already available in the results of the `ec2.DescribeInstances` call.

So this PR also:

- Combines the lookup of Node names, ASG names, and tags into one function, `getNodeInfo`, which replaces the former `retrieveNodeName`.
- Folds the `CheckIfManaged` process into `getNodeInfo`:
  -  There is a new Boolean option, `--assume-asg-tag-propagation`. If `true`, `getNodeInfo` will read the ASG name and the `ManagedAsgTag` directly from the tags in the instance record, and if they are not there it will assume that the instance has no ASG and/or is not managed -- it will not make calls to the `autoscaling` API.
  - `--assume-asg-tag-propagation` defaults to `false`, in which case NTH behaves as before and calls the `autoscaling` API to look up the ASG and its tags. This is to avoid breaking backward compatibility for any NTH users who do not have Tag Propagation set.  

Notes:
- Only as I write this do I realize that there may be users of NTH who run multiple NTH processes, with different `ManagedAsgTag` settings, targeting a single SQS queue. That would be an extremely inefficient strategy because the NTH processes will grab each other's events and delay them by one or more visibility timeouts, but that doesn't mean people aren't doing it. So does the feature which immediately drops messages that fail the `CheckIfManaged` needs to be a configurable option which defaults to the current (slow) behavior? 
 
- I left in the code which, in cases where an instance does not have an `aws:autoscaling:groupName` tag, assumes that the instance might nonetheless belong to an ASG and calls the `autoscaling` API to make sure. I'm not sure these code paths need to exist. Is that `aws:autoscaling:groupName` tag a Sufficiently Official AWS Feature that we can get rid of that logic and just assume the tag will always be there?

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -307,7 +307,7 @@ func (c Config) PrintHumanConfigArgs() {
 			"\tqueue-url: %s,\n"+
 			"\tcheck-asg-tag-before-draining: %t,\n"+
 			"\tmanaged-asg-tag: %s,\n"+
-			"\assume-asg-tag-propagation: %t,\n"+
+			"\tassume-asg-tag-propagation: %t,\n"+
 			"\taws-endpoint: %s,\n",
 		c.DryRun,
 		c.NodeName,


bwagner5:
Ah I believe it's just a "mock" in the test where we're using localstack instead of the real EC2 APIs and the instance is not tagged with the ASG system tag here https://github.com/aws/aws-node-termination-handler/blob/main/test/e2e/asg-lifecycle-sqs-test#L45 .  

I believe adding this to those run-instances calls in the tests for all the files ending in `-sqs-test` in the `test/e2e` directory will fix this: 

```
awslocal ec2 run-instances --private-ip-address ${WORKER_IP} --region ${AWS_REGION} --tag-specifications 'ResourceType=instance,Tags=[{Key=aws:autoscaling:groupName,Value=nth-integ-test}]'
```

PR #539, opened 2021-11-24
Add configurable taint effect

### LINKED ISSUE #273: Configurable taint effect
It would be great if the taint applied to nodes when they are due to terminate could be configured as `NoExecute`.

Currently the taint is only really useful for informational purposes as cordoning the node already applies a `NoSchedule` taint.

I'd like to be able to set terminating nodes as `NoExecute` in order to stop daemonset pods gracefully (specifically to allow Consul agents to leave the cluster)
Similar to https://github.com/kubernetes/kubernetes/issues/75482

hamishforbes:
Add `TAINT_EFFECT` env var and `--taint-effect` flag to configure the node taint effect
Defaults to `NoSchedule`, garbage values will log a warning and default to `NoSchedule`.

closes  #273

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -67,6 +67,8 @@ const (
 	metadataTriesDefault                    = 3
 	cordonOnly                              = "CORDON_ONLY"
 	taintNode                               = "TAINT_NODE"
+	taintEffectDefault                      = "NoSchedule"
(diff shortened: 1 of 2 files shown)


snay2 (approved):
LGTM, thank you for implementing this!

PR #555, opened 2022-01-03
Wrap volumes key in conditional to protect from being nil

gaffneyd4:
Description of changes:

If `webhookTemplateConfigMapName` and `webhookTemplateConfigMapKey` are not set in the values.yaml file, we end up with a deployment.yaml file that has key `volumes` with a `nil` value. This causes other tools like `kustomize` to throw warnings:

```
2022/01/03 18:04:29 nil value at `volumes.configMap.name` ignored in mutation attempt
2022/01/03 18:04:29 nil value at `volumes.projected.sources.configMap.name` ignored in mutation attempt
2022/01/03 18:04:29 nil value at `volumes.secret.secretName` ignored in mutation attempt
2022/01/03 18:04:29 nil value at `volumes.projected.sources.secret.name` ignored in mutation attempt
2022/01/03 18:04:29 nil value at `volumes.persistentVolumeClaim.claimName` ignored in mutation attempt
```

This PR wraps the `volumes` key in the conditional as well so that if it is not being used it doesn't cause these warnings.

### DIFF
--- a/config/helm/aws-node-termination-handler/templates/deployment.yaml
+++ b/config/helm/aws-node-termination-handler/templates/deployment.yaml
@@ -26,12 +26,12 @@ spec:
         {{ $key }}: {{ $value | quote }}
       {{- end }}
     spec:
+      {{- if and .Values.webhookTemplateConfigMapName .Values.webhookTemplateConfigMapKey }}
       volumes:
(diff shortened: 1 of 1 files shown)


bwagner5 (approved):
Awesome! Thank you!

PR #556, opened 2022-01-04
Initial NTH v2 design proposal

bwagner5:
Issue #, if available:
N/A

Description of changes:
 - Adds a design proposal for NTH v2

To view with markdown rendering: https://github.com/bwagner5/aws-node-termination-handler/blob/v2-proposal/designs/v2.md

### DIFF
--- a/designs/v2.md
+++ b/designs/v2.md
@@ -1,0 +1,94 @@
+# AWS Node Termination Handler v2
+*Authors: bwagner5@*
+
+---
+Definitions
+
+* NTH — (AWS) Node Termination Handler
+* IMDS — (EC2) Instance MetaData Service
+* k8s — Kubernetes
+* CRD - Custom Resource Definitions
+* Spot ITN - Spot Interruption Termination Event
+---
+
+## Mission Statement
+
+Simplify termination handling of nodes in Kubernetes clusters.
+
+## Background
+
+NTH has evolved from a simple k8s DaemonSet, now known as IMDS mode, to its current 
+iteration of a Deployment that acts as a queue consumer of EventBridge interruption 
+events. There are pros and cons to selecting which mode to use, but in general IMDS 
+mode is simple and handles Spot ITNs. The queue-processor mode of NTH is 
+more difficult to setup, but is also more powerful since it is able to handle ASG 
+termination events, EC2 instance status changes, Spot Rebalance Recommendations, Spot ITNs, and more.
+
+## Goals
+
+* Simplify the runtime configuration of NTH
+* Support of custom drain events 
+* Custom and configurable actions to drain events
+* Integration w/ Provisioners like Karpenter
+
+## Overview
+
+NTH Queue-Processor mode has accumulated a wealth of configuration options which can make it cumbersome to install.
+Configurations options and different modes of operation have caused the Helm Chart to become large and complex.
+
+NTH solves two problems with one tool: 
+  1. Termination Handling (cordon + drain) 
+  2. Interruption Handling (Spot ITNs, ASG Scale-In, etc.). 
+  
+**Termination Handling** is a fairly simple operation that NTH should handle well and allow for its logic to be plugged into other systems that require 
+out-of-band termination handling.
+
+**Interruption Handling** in v1 only works with a subset of interruption events that are provided by EventBridge or IMDS.
+With NTH v2, custom interruption events should be possible via a labeling mechanism or a custom event monitor. 
+In addition, the actions taken by the interruption monitor should be customizable within a CRD definition.
+
+## The Terminator CRD
+
+To support an extensible system, NTH v2 will use a CRD called a `Terminator` which will represent a logical Node Termination Handler. 
+The Terminator CRD will define which nodes it is in charge of, which interruption events it will monitor, and what actions to take when 
+an interruption event is observed.
+
+NTH v2 will not use helm for configuration of termination options and will instead use the Terminator CRD. This will allow for easier tuning of the 
+runtime configuration as well as the ability to create logical Terminators rather than deploying NTH multiple times which can waste compute resources 
+and is tedious to manage. 
+
+Logical Terminators allow users to customize termination of specific node groups or provisioner managed nodes using a node selector. For example, a 
+logical terminator could be configured to respond to events on nodes with the label `training-group-1`.  Nodes in `training-group-1` may need to some 
+extra time to drain, which can be configured on the Terminator resource in the `pod-termination-grace-period` and the `node-termination-grace-period`. 
+Other groups could use a second logical Terminator resource with more aggressive grace periods.
+
+### Example Terminator:
+
+```
+apiVersion: terminator.k8s.aws/v1alpha1
+kind: Terminator
+metadata:
+  name: default
+spec:
+  matchLabels:
+     node.k8s.aws/capacity-type: spot 
+  queueURL: https://sqs.us-east-42.amazonaws.com/my-nth-q
+  events:
+     SPOT_ITN: ["CORDON", "DRAIN"]
+     SPOT_REBALANCE: ["NO_ACTION"]
+     ASG_TERMINATION: ["CORDON", "DRAIN"]
+     CUSTOM_EVENT1: ["tag:terminate-me"]
+  drainConfig:
+     ignoreDaemonSets: true
+     deleteLocalData: true
+     podTerminationGracePeriod: 30
+     nodeTerminationGracePeriod: 90
+  webhook:
+     url: mywebhook.example.com
+     proxy: myproxy.example.com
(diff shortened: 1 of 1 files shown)


designs/v2.md
```diff
@@ -0,0 +1,94 @@
...
+*Authors: bwagner5@*
+
+---
+Definitions
+
+* NTH — (AWS) Node Termination Handler
+* IMDS — (EC2) Instance MetaData Service
+* k8s — Kubernetes
+* CRD - Custom Resource Definitions
+* Spot ITN - Spot Interruption Termination Event
```
brycahta, line 11: nit: `v1 - Current/existing implementation of NTH`

designs/v2.md
```diff
@@ -0,0 +1,94 @@
...
+* Simplify the runtime configuration of NTH
+* Support of custom drain events 
+* Custom and configurable actions to drain events
+* Integration w/ Provisioners like Karpenter
+
+## Overview
+
+NTH Queue-Processor mode has accumulated a wealth of configuration options which can make it cumbersome to install.
+Configurations options and different modes of operation have caused the Helm Chart to become large and complex.
+
```
brycahta, line 38: nit: example of helm installation step with a realistic config of nth

designs/v2.md
```diff
@@ -0,0 +1,94 @@
...
+     nodeTerminationGracePeriod: 90
+  webhook:
+     url: mywebhook.example.com
+     proxy: myproxy.example.com
+     headers:
+        - context-type: application/json
+     template: |
+        {"text":"[NTH][Instance Interruption] EventID: {{ .EventID }} - Kind: {{ .Kind }} - Instance: {{ .InstanceID }} - Node: {{ .NodeName }} - Description: {{ .Description }} - Start Time: {{ .StartTime }}"}
+...
+```
```
brycahta, line 94: related to comment above, a comparison of an equivalent QP/IMDS NTH config to this would make some of the benefits clearer

brycahta (commented):
Big fan of the approach 👍

designs/v2.md
```diff
@@ -0,0 +1,94 @@
...
+an interruption event is observed.
+
+NTH v2 will not use helm for configuration of termination options and will instead use the Terminator CRD. This will allow for easier tuning of the 
+runtime configuration as well as the ability to create logical Terminators rather than deploying NTH multiple times which can waste compute resources 
+and is tedious to manage. 
+
+Logical Terminators allow users to customize termination of specific node groups or provisioner managed nodes using a node selector. For example, a 
+logical terminator could be configured to respond to events on nodes with the label `training-group-1`.  Nodes in `training-group-1` may need to some 
+extra time to drain, which can be configured on the Terminator resource in the `pod-termination-grace-period` and the `node-termination-grace-period`. 
+Other groups could use a second logical Terminator resource with more aggressive grace periods.
```
snay2, line 63: Could be useful to show both of these configurations in the example below, to make it more clear how the different node groups are selected and handled.

stevehipwell:
This approach looks great but have you considered splitting the monitoring and termination workloads?

A rough suggestion would be a new project containing a termination controller and CRDs for termination events and termination configuration. The controller would watch for termination event CRDs being created and process them with the relevant termination configuration. The termination controller would be abstracted and wouldn't be tied to any specific implementation for creating the termination events. The AWS NTH v2 could then just be a producer of termination events, as could Karpenter.

Alternatively, with the same logic as above, the AWS NTH v2 could combine monitor controller and termination controller which would still support Karpenter.

bwagner5:
> This approach looks great but have you considered splitting the monitoring and termination workloads?
> 
> A rough suggestion would be a new project containing a termination controller and CRDs for termination events and termination configuration. The controller would watch for termination event CRDs being created and process them with the relevant termination configuration. The termination controller would be abstracted and wouldn't be tied to any specific implementation for creating the termination events. The AWS NTH v2 could then just be a producer of termination events, as could Karpenter.
> 
> Alternatively, with the same logic as above, the AWS NTH v2 could combine monitor controller and termination controller which would still support Karpenter.

We've definitely thought about breaking out the Interruption Monitoring logic and Termination Handling logic to support custom use-cases like you're talking about. I don't think it would make sense from a maintainability perspective to separate the two into separate projects, but I definitely agree that they need to be separated logically into separate controllers and be very loosely coupled from one another. 

The Termination CRD is an interesting idea. I was thinking this would be a little heavy for the use-case and was thinking more along the lines of applying a label to the node (this is how Karpenter currently operates). The Termination CRD may give a looser coupling from Interruption Events to Termination since the CRD could contain the exact drain configuration and node set for a workflow style drain (ie handle PDBs within the drain group, potentially drain LoadBalancers, support AMI upgrade strategies, etc).

stevehipwell:
> We've definitely thought about breaking out the Interruption Monitoring logic and Termination Handling logic to support custom use-cases like you're talking about. I don't think it would make sense from a maintainability perspective to separate the two into separate projects, but I definitely agree that they need to be separated logically into separate controllers and be very loosely coupled from one another.

@bwagner5 would this enable NTH to support multiple monitoring controllers and maybe be cloud agnostic?

> The Termination CRD is an interesting idea. I was thinking this would be a little heavy for the use-case and was thinking more along the lines of applying a label to the node (this is how Karpenter currently operates). The Termination CRD may give a looser coupling from Interruption Events to Termination since the CRD could contain the exact drain configuration and node set for a workflow style drain (ie handle PDBs within the drain group, potentially drain LoadBalancers, support AMI upgrade strategies, etc).

I was thinking a CRD would make it easier to handle termination lifecycles; it would allow multiple termination events against a single node with a status for the termination to be captured and K8s events to be generated. I was still expecting the termination logic to come from a separate CRD and mapped from the termination CRD (e.g. SPOT vs NODE_REFRESH).

bwagner5:
> @bwagner5 would this enable NTH to support multiple monitoring controllers and maybe be cloud agnostic?

Yes!

> I was thinking a CRD would make it easier to handle termination lifecycles; it would allow multiple termination events against a single node with a status for the termination to be captured and K8s events to be generated. I was still expecting the termination logic to come from a separate CRD and mapped from the termination CRD (e.g. SPOT vs NODE_REFRESH).

I think we're on the same page here. The setup of interruption events to the termination actions would still be in the Terminator CRD (main NTH CRD). The Termination CRD would contain the exact termination process defined for the workflow from the Terminator CRD.  We can explore how this works in more detail as we get a more detailed design document.

stevehipwell:
Just a couple of further thoughts and obviously the actual design still needs to be done, but using a termination CRD named for the node would allow termination triggers to be aggregated and the status of the termination to be captured and monitored. It'd also be great if the terminator CRD could be selected based on the termination CRD metadata (e.g. termination type) and node metadata (e.g. labels). The following termination patterns could then be supported and combined.

- Default
  - Cordon node (could support taint not cordon)
  - Terminate pods with grace period
  - Terminate node after all pods stopped or grace period
- Cordon and wait
  - Cordon node so no new workloads can be scheduled (could support taint not cordon)
  - Wait for a set period or until all pods have finished
  - Pods to wait for could be matched with an expression
  - If pods are still running at the end of the period fall back to another (or default) pattern
- Schedule
  - Wait until a specific time before starting another pattern

As an example a capacity rebalance termination trigger might start a cordon and wait pattern but if a terminate immediately trigger was added for that node then the cordon and wait could be skipped and the default pattern attempted immediately. Equally it should be possible to stop a termination by flagging the CRD as cancelled (or deleting it).

stevehipwell:
@bwagner5 has there been any progress on the designs for v2?

sauryadas:
Couple of comments-
1) Will NTH v2 work with spot managed node groups?
2) How will this behave if k8s PDB's are defined ?

bwagner5:
> Couple of comments-
> 
> 1. Will NTH v2 work with spot managed node groups?

NTH is not necessary on EKS spot managed node groups, because they have their own managed termination handler. 
https://github.com/aws/aws-node-termination-handler/discussions/559

> 2. How will this behave if k8s PDB's are defined ?

PDBs are respected with NTH v1 and will continue to w/ v2 since we are using the k8s eviction API.

designs/v2.md
```diff
@@ -0,0 +1,94 @@
...
+metadata:
+  name: default
+spec:
+  matchLabels:
+     node.k8s.aws/capacity-type: spot 
+  queueURL: https://sqs.us-east-42.amazonaws.com/my-nth-q
+  events:
+     SPOT_ITN: ["CORDON", "DRAIN"]
+     SPOT_REBALANCE: ["NO_ACTION"]
+     ASG_TERMINATION: ["CORDON", "DRAIN"]
```
ellistarn, line 79: Why would a user want to configure these? Are there sensible defaults they can use instead?

PR #565, opened 2022-01-31
🥑🤖 v1.14.1 release prep

snay2:
🥑🤖 Auto-generated PR for v1.14.1 release. Updating release versions in repo.

### DIFF
--- a/config/helm/aws-node-termination-handler/Chart.yaml
+++ b/config/helm/aws-node-termination-handler/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: aws-node-termination-handler
 description: A Helm chart for the AWS Node Termination Handler.
 type: application
-version: 0.16.0
-appVersion: 1.14.0
+version: 0.16.1
+appVersion: 1.14.1
 kubeVersion: ">= 1.16-0"
 keywords:
   - aws


brycahta (approved):
/lgtm

stevehipwell:
@brycahta @snay2 @bwagner5 as #540 contained breaking changes shouldn't the chart version have been updated to `0.17.0`? I know that technically any change to a SemVer 2 `v0.x.y` version is a breaking change but it's clearer if the minor version is changed.

snay2:
@stevehipwell Thanks for bringing that to our attention. In hindsight, I should have made this a minor version release (`0.17.0`) instead of a patch release because there were technically breaking changes. But after reviewing it with the team this morning, we feel comfortable leaving the helm chart at `0.16.1` despite those changes.

snay2:
On second thoughts, since the chart hasn't been released to eks-charts yet (https://github.com/aws/eks-charts/pull/675), we're going to take this opportunity to match the version numbers between the chart and the app, and we won't release a chart on version `0.16.1`.

Therefore, the breaking changes for the helm chart refactor will be in the chart version `1.14.1` to match the NTH app version. PR forthcoming...

stevehipwell:
I'd advise against trying to match the chart and app versions, how about chart `v1.0.0`?

PR #567, opened 2022-01-31
use gha cache

bwagner5:
Issue #, if available:
N/A

Description of changes:
 - Cache go mods

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,15 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
+    - name: Restore go mod cache
+      uses: actions/cache@v2
+      with:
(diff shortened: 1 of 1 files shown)


snay2 (commented):
What is the intention of this change? Should it improve build times? Reduce build failures?

Does it come with any additional costs (e.g., GHA capacity usage)?

snay2 (approved):
Discussed during the SIG meeting. Risks are low. Build times improved dramatically when they did this on Karpenter, so we're going to merge it and see what benefit we get.

bwagner5:
I think I remember why I haven't done this before... because we're building NTH in containers so we'd need to do some storage mapping to get this to work properly.... I'm going to merge since this will still be necessary and doesn't hurt things currently, but I guess there's some more work to do here :)

Issue #356, opened 2021-01-28
Support ASG to SQS


foriequal0:
It seems that ASG LifecycleHook can directly send events to SQS: (See NotificationTargetARN https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_LifecycleHook.html)

Is there any special reason to use EventBridge?

bwagner5:
Good point!  Unfortunately, the current implementation does not support the payload body of ASG sending it directly, since it's different from EventBridge.  I think it would be an excellent feature to support though. 

```
{
  "LifecycleHookName": "test-to-sqs",
  "AccountId": "000000000000",
  "RequestId": "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxx",
  "LifecycleTransition": "autoscaling:EC2_INSTANCE_TERMINATING",
  "AutoScalingGroupName": "my-asg,
  "Service": "AWS Auto Scaling",
  "Time": "2021-01-28T20:06:18.130Z",
  "EC2InstanceId": "i-00000000000",
  "LifecycleActionToken": "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxx"
}
```

patstrom:
Is it only the data format that differs? Or are the other architectural concerns?

I am looking to deploy this project in the queue-processor mode and it would be nice to do it with as little complexity as possible. If EventBridge isn't strictly necessary I would prefer to not utilize (and pay for) it. Depending on how big of a patch this would require I could give it a go.

foriequal0:
It seems that EventBridge exists to funnel all events in the account into a single global queue channel.
ASG to SQS only sends two kind of events: `autoscaling:EC2_INSTANCE_LAUNCHING`, `autoscaling:EC2_INSTANCE_TERMINATING`. No Spot ITN, no scheduled events, no rebalance recommendation with ASG to SQS alone, and it seems that there is no way to listen EC2 status changes.
To use it alone without EventBridge, we should make a IMDS+Queue hybrid processor. I think it'll be more complicated.

JnMik:
Hmm.. I guess that must be why I have these errors _ignoring event due to error error="Event source () is not supported_ : 
![image](https://user-images.githubusercontent.com/4965516/118558696-016cf680-b735-11eb-8b05-02056f4a9662.png)

I have ASG > Terminate hook > SQS and I've passed the SQS Queue URL to the NTH (v1.13.0)

Sreddy-gs:
[...]

PR #568, opened 2022-02-01
Support ASG-to-SQS Lifecycle Termination events

AustinSiu:
Issue #, if available: #356

Description of changes:

* Check SQS message for a Lifecycle termination event sent directly from ASG to SQS (different schema than from EventBridge to SQS)
* Update README with new/updated instructions to set up Lifecycle hook and IAM access roles

These changes are meant to help support only Lifecycle termination events from ASG sent to SQS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,6 +288,8 @@ See all the different events docs [here](https://docs.aws.amazon.com/eventbridge
 
 #### 4. Create Amazon EventBridge Rules
 
+You may skip this step if sending events from ASG to SQS directly.
+
 Here are AWS CLI commands to create Amazon EventBridge rules so that ASG termination events, Spot Interruptions, Instance state changes, Rebalance Recommendations, and AWS Health Scheduled Changes are sent to the SQS queue created in the previous step. This should really be configured via your favorite infrastructure-as-code tool like CloudFormation or Terraform:
 
 ```
@@ -326,8 +328,6 @@ $ aws events put-rule \
 $ aws events put-targets --rule MyK8sScheduledChangeRule \
   --targets "Id"="1","Arn"="arn:aws:sqs:us-east-1:123456789012:MyK8sTermQueue"
 ```
-
-You may skip this step if sending events from ASG to SQS directly.
 
 #### 5. Create an IAM Role for the Pods
 
--- a/README.md
+++ b/README.md
@@ -212,6 +212,17 @@ $ aws autoscaling put-lifecycle-hook \
   --heartbeat-timeout=300
 ```
 
+If you want to avoid using EventBridge and instead send ASG Lifecycle events directly to SQS, you will need to first create the SQS queue and role to grant access permissions. Please follow the next steps until you've completed Step 3, then come back to this step and use the following AWS CLI command, using the ARNs from Step 3:
+
+```
+$ aws autoscaling put-lifecycle-hook \
+  --lifecycle-hook-name=my-k8s-term-hook \
+  --auto-scaling-group-name=my-k8s-asg \
+  --lifecycle-transition=autoscaling:EC2_INSTANCE_TERMINATING \
+  --notification-target-arn <your test queue ARN here> \
+  --role-arn <your SQS access role ARN here>
+```
+
 #### 2. Tag the ASGs:
 
 By default the aws-node-termination-handler will only manage terminations for ASGs tagged w/ `key=aws-node-termination-handler/managed`
@@ -269,6 +280,9 @@ EOF
 
 $ aws sqs create-queue --queue-name "${SQS_QUEUE_NAME}" --attributes file:///tmp/queue-attributes.json
 ```
+
+If you are sending Lifecycle termination events from ASG directly to SQS, instead of through EventBridge, then you will also need to create an IAM service role to give Amazon EC2 Auto Scaling access to your SQS queue. Please follow [these linked instructions to create the IAM service role: link.](https://docs.aws.amazon.com/autoscaling/ec2/userguide/configuring-lifecycle-hook-notifications.html#sqs-notifications)
+Note the ARNs for the SQS queue and the associated IAM role, and return to Step 1 at this time to set up the direct-from-ASG-to-SQS Lifecycle hook.
 
 #### 4. Create Amazon EventBridge Rules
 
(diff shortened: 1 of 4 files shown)


pkg/monitor/sqsevent/sqs-monitor.go
```diff
@@ -101,6 +101,19 @@ func (m SQSMonitor) processSQSMessage(message *sqs.Message) (*EventBridgeEvent,
 	event := EventBridgeEvent{}
 	err := json.Unmarshal([]byte(*message.Body), &event)
 
+	if err == nil && len(event.DetailType) == 0 && len(event.Time) > 0 {
```
bwagner5, line 104: can you conform this to the `if err != nil` style. I think it's worth the explicitness and reading left alignment as the success case instead of saving a few lines of code. 

Here's the general principle: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

bwagner5 (changes requested):
nice work! just a few comments

README.md
```diff
@@ -270,6 +281,9 @@ EOF
 $ aws sqs create-queue --queue-name "${SQS_QUEUE_NAME}" --attributes file:///tmp/queue-attributes.json
 ```
 
+If you are sending Lifecycle termination events from ASG directly to SQS, instead of through EventBridge, then you will also need to create an IAM service role to give Amazon EC2 Auto Scaling access to your SQS queue. Please follow [these linked instructions to create the IAM service role: link.](https://docs.aws.amazon.com/autoscaling/ec2/userguide/configuring-lifecycle-hook-notifications.html#sqs-notifications)
```
brycahta, line 284: We should aim to be less verbose and with fewer redirects wrt documentation and setup. Is it possible to re-work these setup steps so users don't need to bounce between them? SQS + IAM setup stand out to me as constants so maybe start with those and deviate from there (whether or not users want EventBridge)? OR start with EventBridge and let steps flow from there.

Applies to above entry as well

pkg/monitor/sqsevent/sqs-monitor.go
```diff
@@ -101,6 +101,19 @@ func (m SQSMonitor) processSQSMessage(message *sqs.Message) (*EventBridgeEvent,
...
+		log.Info().Msg("no EventBridge detail-type found, assuming a Lifecycle termination event")
```
brycahta, line 105: why is `len(event.Time) > 0` also required?

pkg/monitor/sqsevent/sqs-monitor.go
```diff
@@ -101,6 +101,19 @@ func (m SQSMonitor) processSQSMessage(message *sqs.Message) (*EventBridgeEvent,
...
+		lifecycleEvent := LifecycleDetail{}
+		err = json.Unmarshal([]byte(*message.Body), &lifecycleEvent)
+
+		if err == nil {
```
brycahta, line 109: nit: can you note that this is modifying event to conform to EventBridgeEvent OR that this msg doesn't come from EventBridge

pkg/monitor/sqsevent/sqs-monitor_test.go
```diff
@@ -87,10 +97,10 @@ func TestKind(t *testing.T) {
 	h.Assert(t, sqsevent.SQSMonitor{}.Kind() == sqsevent.SQSTerminateKind, "SQSMonitor kind should return the kind constant for the event")
 }
 
-func TestMonitor_Success(t *testing.T) {
+func TestMonitor_EventBridgeSuccess(t *testing.T) {
 	spotItnEventNoTime := spotItnEvent
 	spotItnEventNoTime.Time = ""
-	for _, event := range []sqsevent.EventBridgeEvent{spotItnEvent, asgLifecycleEvent, spotItnEventNoTime, rebalanceRecommendationEvent} {
+	for _, event := range []sqsevent.EventBridgeEvent{spotItnEvent, asgLifecycleEventFromEventBridge, spotItnEventNoTime, rebalanceRecommendationEvent} {
```
brycahta, line 103: nit: `spotItnEvent` is also from EventBridge

brycahta (changes requested):

pkg/monitor/sqsevent/sqs-monitor_test.go
brycahta: after looking, can you just update this name to `asgLifecycleEvent`?

pkg/monitor/sqsevent/sqs-monitor.go
AustinSiu: Originally added since it is a common field between EventBridge and Lifecycle events, and caused some confusing behavior with handling EventBridge events with default values. However, after reviewing again, I'll remove this in favor of more explicit checks for Lifecycle events.

brycahta (dismissed):
/lgtm looks clean 👍

README.md
```diff
@@ -311,6 +327,8 @@ $ aws events put-targets --rule MyK8sScheduledChangeRule \
   --targets "Id"="1","Arn"="arn:aws:sqs:us-east-1:123456789012:MyK8sTermQueue"
 ```
 
+You may skip this step if sending events from ASG to SQS directly.
```
snay2, line 330: Love the reordered steps, they make the setup process much more clear!

This sentence would be better at the beginning of step 4, so the customer doesn't have to read any more than necessary.

pkg/monitor/sqsevent/sqs-monitor.go
```diff
@@ -101,9 +101,37 @@ func (m SQSMonitor) processSQSMessage(message *sqs.Message) (*EventBridgeEvent,
 	event := EventBridgeEvent{}
 	err := json.Unmarshal([]byte(*message.Body), &event)
 
+	if err != nil {
+		return &event, err
+	}
+
+	if len(event.DetailType) == 0 {
+		log.Debug().Msg("processing message from source other than EventBridge")
```
snay2, line 109: Do we know for sure that this event is from ASG, or could it be something else? If the former, can we make the debug message more specific? If the latter, does the function `processLifecycleEventFromASG()` need a more generic name?

bwagner5 (approved):
lgtm

brycahta (approved):
/lgtm

PR #570, opened 2022-02-02
allow for envFrom on helm chart

danquack:
Description of changes:
We store a handful of sensitive values in secrets, so this just allows for the envFrom instead of a large array of extraEnv.


config/helm/aws-node-termination-handler/values.yaml
```diff
@@ -54,6 +54,7 @@ tolerations: []
 
 # Extra environment variables
 extraEnv: []
+envFrom: []
```
snay2, line 57: Can you also add this to the helm chart README? https://github.com/aws/aws-node-termination-handler/blob/main/config/helm/aws-node-termination-handler/README.md?plain=1#L70

snay2 (changes requested):

config/helm/aws-node-termination-handler/README.md
```diff
@@ -68,6 +68,7 @@ The configuration in this table applies to all AWS Node Termination Handler mode
...
 | `extraEnv`                         | Additional environment variables for the _aws-node-termination-handler_ container.                                                                                                                                                                                                                                                                                                     | `[]`                                                  |
+| `envFrom`                          | Additional environment variables for the _aws-node-termination-handler_ container.                                                                                                                                                                                                                                                                                                     | `[]`
```
snay2, line 71: Thanks for adding this. Could you update the description for one or both of the `extraEnv` and `envFrom` parameters to make clear the distinction between them? Some things to consider may be where the parameters are used in the chart and whether their syntax is different.

snay2 (changes requested):

stevehipwell:
@danquack do you think it's worth adding `envFrom` instead of using the secret reference directly from `env`? I'm not sure I have a strong opinion but worth considering?

Issue #316, opened 2020-12-11
Handle ELB instance deregistration


sarahhodne:
We've noticed in our production environment that we have a need for something to deregister nodes from load balancers as part of the draining procedure, before the instance is terminated. We're currently using [lifecycle-manager](https://github.com/keikoproj/lifecycle-manager) for this, but it would be nice if this was handled by the AWS Node Termination Handler instead.

The reason this is needed is that if the instance is terminated before it's deregistered from an ELB, a number of connections will fail until the health check starts failing. This is particularly noticeable on ELBv2 (NLB+ALB), which seem to take several minutes to react, so we need to have fairly high timeout times on the health checks.

The behaviour we're looking for is that the node termination handler finds a list of classic ELBs and target groups that it's a member of, sends a deregistration request and then waits for the deregistration to finish before marking the instance as being ready to terminate.

yuri-1987:
same issue here, in addition, when termination handler cordons node, the node marked as `unschedulable`, 
the service controller removes cordoned nodes from LB pools, it can potentially drop in-flight requests, there should be a better process for a node draining,
1. taint the node (don't cordon)
2. find elb's/target groups and safely de-register the node
3. cordon
4. drain

relevant issues:
https://github.com/kubernetes/autoscaler/issues/1907
https://github.com/kubernetes/kubernetes/issues/65013
https://github.com/kubernetes/kubernetes/issues/44997

and a partial bug fix in 1.19
https://github.com/kubernetes/kubernetes/pull/90823

bwagner5:
I'm definitely interested in looking into this more. I've asked @kishorj who works on the aws-load-balancer-controller his thoughts since there needs to be a careful dance between the LB controller and NTH in the draining process. There might be more we can do in that controller without involving NTH as much. But if we need to add this logic to NTH, then I'm not opposed

yuri-1987:
Hi Brandon, thank you for the quick response. I think an external tool such as NTH is suitable to handle such logic. Even if Kubernetes contributors solve it internally, it won't manage all cases such as draining due to spot interruptions, AZ rebalance, or spot recommendations. The current bug of removing cordon nodes immediately from the load balancer is four years old, if the service controller will be enhanced someday, it can take a lot of time till we can use it. I really hope to see this functionality in NTH.

bwagner5:
Linking taint-effect issue, since I think that would mitigate this: https://github.com/aws/aws-node-termination-handler/issues/273

[...]

PR #582, opened 2022-02-10
Add and remove exclude-from-external-load-balancers label

tjs-intel:
Issue #, if available: https://github.com/aws/aws-node-termination-handler/issues/316

Description of changes:
Apply the label `node.kubernetes.io/exclude-from-external-load-balancers=aws-node-termination-handler` to tell load balancer controllers (like `aws-load-balancer-controller`) to exclude the node from load balancers before it is brought offline. If the label is already applied, don't do anything. When uncordoning the node, check that the label's value is `aws-node-termination-handler` before removing it.


tjs-intel:
Tested manually on an EKS cluster:
 * Observed the label being added successfully when decreasing the desired capacity of the ASG (also observed the associated target draining).
 * Applied the label `node.kubernetes.io/exclude-from-external-load-balancers=true` to a node, triggered a downscale (reduced desired capacity) and observed the value of the label to be unchanged.

I have not tested the label removal when uncordoning a node.

pkg/node/node.go
```diff
@@ -282,6 +315,41 @@ func (n Node) removeLabel(nodeName string, key string) error {
...
+	}
+	payload, err := json.Marshal(append(patchReqs, patchRemove))
+	if err != nil {
+		return fmt.Errorf("An error occurred while marshalling the json to remove a label from the node: %w", err)
+	}
+	node, err := n.fetchKubernetesNode(nodeName)
+	if err != nil {
+		return err
+	}
+	if n.nthConfig.DryRun {
```
snay2, line 338: We need to check whether the value matches (line 343) before doing this. Otherwise, during dry-runs we will always print out "Would have removed label", even when the value doesn't match and thus we would not have removed the label during actual operation.

Can you move this block to right before line 346?

snay2 (changes requested):

pkg/node/node.go
tjs-intel: Can do!

bwagner5:
Would there be any harm in always labeling with the `node.kubernetes.io/exclude-from-external-load-balancers` label?

We have so much configuration already, just curious if it makes sense to put it behind a flag or if we should always add the label as part of a graceful termination.  I'm not strongly opposed to a flag, so if there is a good reason why someone would not want to have a node labeled with this, then this LGTM!

bwagner5 (approved):
Lgtm!

snay2 (approved):

tjs-intel:
These failing tests seem to be permission errors, hopefully nothing I'm responsible for causing!

snay2:
@tjs-intel Looks like it was a transient failure; they passed when I re-ran them.

PR #571, opened 2022-02-02
allow for dnsConfig directive

danquack:
Description of changes: Allow for dnsConfig options on the helm chart: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-dns-config

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #586, opened 2022-02-15
:bug: :fire: only add the dns config not the entire chart....

danquack:
Issue #, if available: Fixes bug introduced in #571

Description of changes: If you set dnsConfig option it put into yaml the whole context...it should have just been the `.Values.dnsConfig`


snay2 (approved):
LGTM! Thanks for fixing this!

PR #587, opened 2022-02-16
replace kustomize with helm

cjerad:
Issue #, if available:

Description of changes:
* Replaced kustomize templates with Helm chart
* Updated Makefile targets


src/Makefile
```diff
@@ -61,59 +58,37 @@ test: manifests generate fmt vet envtest ## Run tests.
...
-kustomize: ## Download kustomize locally if necessary.
-	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v3@v3.8.7)
+KO = $(shell pwd)/bin/ko
+.PHONY: ko
+ko: ## Download ko locally if necessary.
+	$(call go-get-tool,$(KO),github.com/google/ko@v0.9.3)
 
 ENVTEST = $(shell pwd)/bin/setup-envtest
 .PHONY: envtest
 envtest: ## Download envtest-setup locally if necessary.
```
bwagner5, line 90: can you move these tool installs to `hack/toolchain.sh`

bwagner5 (approved):
nice! I just had one comment on moving the tool installs to a `hack/toolchain.sh` script. I don't want the Makefile to get out of control. It should be really simple stuff in it. Feel free to merge this and do in a follow-up PR though.

src/Makefile
```diff
@@ -37,22 +39,26 @@ all: build
...
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
 .PHONY: generate
-generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
+generate:  # Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 .PHONY: fmt
-fmt: ## Run go fmt against code.
+fmt: # Run go fmt against code.
```
bwagner5, line 57: add double `##` back since this is used to denote usage for `make help` rather than a Makefile comment

src/scripts/toolchain.sh
```diff
@@ -0,0 +1,52 @@
...
+    echo "error: missing destination path"
+    echo "$usage" 1>&2
+    exit 1
+fi
+
+tmp_dir="$(mktemp -d)"
+trap "rm -rf \"$tmp_dir\"" EXIT
+
+cd "$tmp_dir"
+go mod init tmp >/dev/null 2>&1
```
bwagner5, line 47: you can remove all of this stuff, `go install` won't update go.mod anymore when using go 1.16 or later.

src/Makefile
```diff
@@ -37,22 +39,26 @@ all: build
...
 	go fmt ./...
 
 .PHONY: vet
-vet: ## Run go vet against code.
+vet: # Run go vet against code.
 	go vet ./...
 
 .PHONY: test
```
bwagner5, line 64: the `setup-envtest` probably needs to go into toolchain. There are some complexities with it like not supporting arm64 macs so you need to set some mock env vars and let Rosetta run it.

we have this in toolchain.sh in karpenter:

```
kubebuilder() {
    mkdir -p $KUBEBUILDER_ASSETS
    arch=$(go env GOARCH)
    ## Kubebuilder does not support darwin/arm64, so use amd64 through Rosetta instead
    if [[ $(go env GOOS) == "darwin" ]] && [[ $(go env GOARCH) == "arm64" ]]; then
        arch="amd64"
    fi
    ln -sf $(setup-envtest use -p path "${K8S_VERSION}" --arch="${arch}" --bin-dir="${KUBEBUILDER_ASSETS}")/* ${KUBEBUILDER_ASSETS}
    find $KUBEBUILDER_ASSETS
}
```

src/Makefile
```diff
@@ -37,22 +39,26 @@ all: build
...
+manifests:  # Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
 .PHONY: generate
-generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
+generate:  # Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 .PHONY: fmt
-fmt: ## Run go fmt against code.
```
bwagner5, line 51: can you merge `fmt` and `vet` into a verify target

bwagner5 (changes requested):

bwagner5 (approved):
lgtm

Issue #589, opened 2022-02-17
GHA displaying error annotations


brycahta:
**Describe the bug**
After **successfully** running `[Build and Test jobs](https://github.com/aws/aws-node-termination-handler/actions/runs/1857963012), Annotations are displays containing errors, warning, and failure. This is confusing cx for contributors and maintainers. 

#### Example Annotations

![image](https://user-images.githubusercontent.com/61433408/154535223-2995eb15-548a-4da5-8a7a-e42e84e2eda9.png)

PR #590, opened 2022-02-17
Move golangci-lint action to its own job

brycahta:
#### Issue #, if available:
* https://github.com/aws/aws-node-termination-handler/issues/589
* https://github.com/golangci/golangci-lint-action/issues/23

#### Description of changes
Moving `golangci-lint` into its own job per the [updated onboarding instructions](https://github.com/golangci/golangci-lint-action#how-to-use)

#### Contribution disclaimer

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -20,7 +20,7 @@ env:
 
 jobs:
   fastTests:
-    name: Fast Tests and Lints
+    name: Fast Test
     runs-on: ubuntu-20.04
     steps:
     - name: Set up Go 1.x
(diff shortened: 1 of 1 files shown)


snay2 (approved):
Do these jobs run in order, and only after the previous jobs complete successfully? If so, it might make sense to put the linter first since it runs faster than the test suite. That would make the feedback loop quicker in the event of a failure.

brycahta:
> Do these jobs run in order, and only after the previous jobs complete successfully?

No, these are concurrent (or at least requests to run are concurrent). The functionality you're describing is achieved using `needs` like in our [release workflow](https://github.com/aws/aws-node-termination-handler/blob/main/.github/workflows/release.yaml#L59): 

```
  release:
    name: Release
    runs-on: ubuntu-20.04
    needs: [releaseLinux, releaseWindows]
    steps:
    - name: Set up Go 1.x
      uses: actions/setup-go@v2
      with:
        go-version: ${{ env.DEFAULT_GO_VERSION }}
```

Issue #584, opened 2022-02-15
Incorrectly pulls amd64 image on arm64 machine.


narlark:
**Describe the bug**
On an arm machine when you pull the multi-arch image, it pulls the amd64 image instead of the arm64 image.

**Steps to reproduce**
Some weird behavior I'm noticing with upstream amazon/aws-node-termination-handler

On inspecting the manifest it's clearly a multi-arch image and does have an arm64 arch version available.
```

$ docker manifest inspect amazon/aws-node-termination-handler:v1.6.1
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 947,
         "digest": "sha256:7e91ba3ff76e3c540f8e2f1d3935b61fb05f831a9c36f42961a5d0e878e7c8a4",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 947,
         "digest": "sha256:f83038b59db9cebe1b1904fc6f351e82f8aa9e6cdb09bd89d34996998ad5ab16",
         "platform": {
            "architecture": "arm",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 947,
         "digest": "sha256:c3e1c2d17a05c8ecb661658bd179be3784fe605da58ab7940e1303c0fb8eda70",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1752,
         "digest": "sha256:2a72849bcd7c46518523072f8ad3453c451030a1ddfd2dd9c95bdae3ffdfdc8e",
         "platform": {
            "architecture": "amd64",
            "os": "windows",
            "os.version": "10.0.17763.1282"
         }
      }
   ]
}
```

But when I pull the image on my arm64 machine, it pulls the amd64 image instead of the arm64 image.
```

$ docker pull amazon/aws-node-termination-handler:v1.6.1

$ docker image inspect amazon/aws-node-termination-handler:v1.6.1
.....

            ],
            "OnBuild": null,
            "Labels": null
        },
        "Architecture": "amd64",
        "Os": "linux",
[...]

PR #594, opened 2022-02-23
Add support for docker buildx on linux builds

snay2:
This ensures that the platform in the image manifest matches the platform of the binaries.

**Issue #, if available:** #584

**Description of changes:**

* Modify the Dockerfile to pass through the docker buildx built-in arguments `TARGETOS` and `TARGETARCH`
* Remove support for `linux/arm` and `darwin/amd64` images, since they are not used in production workloads. Continuing to build them now would require using a different base image than `amazonlinux:2`, which does not support those architectures.
* Continue using the older build system for Windows images, since docker buildx in our current GitHub Actions configuration does not support Windows builds.
* Add a comment that explicitly annotating the image OS and architecture may not be necessary in the future

**Testing:**
Tested with GitHub Actions on my fork, publishing to an ECR Public repo in my personal AWS account. Windows and Linux builds worked successfully, and the manifests are correct.

### DIFF
--- a/scripts/build-docker-images
+++ b/scripts/build-docker-images
@@ -47,6 +47,7 @@ while getopts "dp:r:v:" opt; do
   esac
 done
 
+
 for os_arch in "${PLATFORMS[@]}"; do
     os=$(echo $os_arch | cut -d'/' -f1)
     arch=$(echo $os_arch | cut -d'/' -f2)
@@ -56,13 +57,23 @@ for os_arch in "${PLATFORMS[@]}"; do
     dockerfile="$DOCKERFILE_PATH"
     if [[ $os = "windows" ]]; then
         dockerfile="${dockerfile}.windows"
+        docker build \
+            --file "${dockerfile}" \
+            --build-arg GOOS=${os} \
+            --build-arg GOARCH=${arch} \
+            --build-arg GOPROXY=${GOPROXY} \
+            --tag ${img_tag} \
+            ${REPO_ROOT_PATH}
+    else
+        # Launch a docker buildx instance and save its name so we can terminate it later
+        buildx_instance_name=$(docker buildx create --use)
+        docker buildx build \
+            --load \
+            --file "${dockerfile}" \
+            --build-arg GOPROXY=${GOPROXY} \
+            --tag ${img_tag} \
+            --platform "${os_arch}" \
+            ${REPO_ROOT_PATH}
+        docker buildx rm ${buildx_instance_name}
     fi
-
-    docker build \
-        --file "${dockerfile}" \
-        --build-arg GOOS=${os} \
-        --build-arg GOARCH=${arch} \
-        --build-arg GOPROXY=${GOPROXY} \
-        --tag ${img_tag} \
-        ${REPO_ROOT_PATH}
 done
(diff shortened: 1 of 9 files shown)


brycahta:
> Continue using the older build system for Windows images, since docker buildx in our current GitHub Actions configuration does not support Windows builds.

What kinda effort are we looking at to align the build with Windows? I'd rather not maintain 2 different systems

snay2:
> > Continue using the older build system for Windows images, since docker buildx in our current GitHub Actions configuration does not support Windows builds.
> 
> What kinda effort are we looking at to align the build with Windows? I'd rather not maintain 2 different systems

@brycahta Agreed, but currently it's a big unknown to me. I didn't discover it until yesterday. Have been trying to get a build running on a Windows laptop with buildx but so far no luck.

Documentation on that use case seems kinda scarce, too, oddly. Lots of "cross-compile all your Linux containers from a Windows machine" but very little of "use buildx on Windows to build Windows containers".

Should be possible, but I don't yet have any idea how much work it will be.

scripts/build-docker-images
```diff
@@ -56,13 +57,23 @@ for os_arch in "${PLATFORMS[@]}"; do
...
+            --file "${dockerfile}" \
+            --build-arg GOOS=${os} \
+            --build-arg GOARCH=${arch} \
+            --build-arg GOPROXY=${GOPROXY} \
+            --tag ${img_tag} \
+            ${REPO_ROOT_PATH}
+    else
+        # Launch a docker buildx instance and save its name so we can terminate it later
+        buildx_instance_name=$(docker buildx create --use)
+        docker buildx build \
```
brycahta, line 70: does `buildx` need to be setup on the gha runner or is it already installed?

scripts/push-docker-images
```diff
@@ -62,6 +62,8 @@ if [[ $MANIFEST == "true" ]]; then
   fi
   cat <<< "$(jq '.+{"experimental":"enabled"}' $DOCKER_CLI_CONFIG)" > $DOCKER_CLI_CONFIG
   echo "Enabled experimental CLI features to execute docker manifest commands"
+  # Delete the local version of the manifest so we rely solely on the remote manifest
+  docker manifest rm $IMAGE_REPO:$VERSION || :
```
brycahta, line 66: I feel like there should be a warning before deleting user's local manifest or maybe a way to restore it because I wouldn't expect a `push-docker-images` script to delete anything

brycahta (changes requested):
Can you also update `docker build` commands in [BUILD.md](https://github.com/aws/aws-node-termination-handler/blob/50f0c4de9ceb5c2e2ccec8144fd2ca56a4749647/BUILD.md)? It could also be a good place to add some info regarding Windows using `build` instead of `buildx`

scripts/build-docker-images
snay2: It was already installed. I didn't have to modify the gha runners and they are able to build successfully.

scripts/push-docker-images
snay2: Running this script on a local developer machine is already an edge case to be done with extreme caution. The primary place this will run is in the CI/CD system, where no one is creating manifests by hand anyway.

The local manifest should not be a source of truth at this stage of the process. A local manifest may lack some images that are present in the remote manifest, so it is almost always incorrect to rely on an existing local manifest that was not created by this script.

Unfortunately `docker manifest inspect` doesn't draw a distinction between local manifests and remote manifests, so this is the only way to ensure we rely on the remote manifest as a starting point.

All that said, what kind of warning would you want to see? I'm assuming you're thinking of something that would get the attention of a human user without disrupting automated builds?

brycahta: > Running this script on a local developer machine is already an edge case to be done with extreme caution

Gotcha, thanks for the context. In that case, this is good to go 👍

brycahta:
> and remove passive voice

 🙏  many thanks

brycahta (approved):
/lgtm

Issue #608, opened 2022-03-29
ServiceMonitor label selector is wrong


paalkr:
**Describe the bug**
The ServiceMonitor object selector doesn't refer to the Service object, but to the Deployment object.

**Steps to reproduce**
Deploy nth in Queue Processor mode with `enablePrometheusServer: true` and `serviceMonitor.create: true`
The ServiceMonitor object will have the `spec.selector.matchLabels.app.kubernetes.io/component: deployment` set, hence no endpoints are picked up by Prometheus for scraping.

**Expected outcome**
The ServiceMonitor object should have `spec.selector.matchLabels.app.kubernetes.io/component:` set to {{ .Release.Name }}, wich is in fact the labels used by the service.

https://github.com/aws/aws-node-termination-handler/blob/85739d23b5e2c4184a092833e0619ecad2d9d215/config/helm/aws-node-termination-handler/templates/service.yaml#L7-L8

https://github.com/aws/aws-node-termination-handler/blob/85739d23b5e2c4184a092833e0619ecad2d9d215/config/helm/aws-node-termination-handler/templates/_helpers.tpl#L51

**Environment**
* NTH App Version: 1.16.0
* NTH Mode (IMDS/Queue processor): QP
* Installation method: Helm 0.18.0

jillmon:
Also maybe https://github.com/aws/aws-node-termination-handler/pull/604

PR #604, opened 2022-03-18
Fix service monitor selector

victorboissiere:
Signed-off-by: Victor Boissiere <victor.boissiere@qonto.com>

**Issue #, if available:**

**Description of changes:**
Fix issue on service monitor selector.

Current service monitor selector:
```yaml
  selector:
    matchLabels:
      app.kubernetes.io/component: deployment
      app.kubernetes.io/instance: aws-node-termination-handler
      app.kubernetes.io/name: termination-handler
     ...
```

But the servicelabel  is
```yaml
apiVersion: v1
kind: Service
metadata:
  labels:
    app.kubernetes.io/component: aws-node-termination-handler <== here
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: termination-handler
    ...
```

Fix the issue by ensuring service monitor is target the same labels as the service

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #609, opened 2022-03-29
Fix ServiceMonitor selector

paalkr:
**Issue #, if available:**
Fixes https://github.com/aws/aws-node-termination-handler/issues/608

**Description of changes:**
Add correct label selector to the ServiceMonitor object

### DIFF
--- a/config/helm/aws-node-termination-handler/templates/_helpers.tpl
+++ b/config/helm/aws-node-termination-handler/templates/_helpers.tpl
@@ -82,6 +82,14 @@ app.kubernetes.io/component: daemonset
 {{- end -}}
 
 {{/*
+Selector labels for the ServiceMonitor
+*/}}
+{{- define "aws-node-termination-handler.selectorServiceMonitor" -}}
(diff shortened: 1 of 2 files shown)


paalkr:
Wow, how did I miss the existing PR https://github.com/aws/aws-node-termination-handler/pull/604 fixing the exact same issue, though in a slightly different way.

paalkr:
Hi @brycahta,

I don't have a very strong feeling about which approach is better. I considered the same solution as in https://github.com/aws/aws-node-termination-handler/pull/604, but went with the new helper function as I think that is more in line with how the podMonitor selector for the Deployment / Daemonset is constructed.
https://github.com/aws/aws-node-termination-handler/blob/85739d23b5e2c4184a092833e0619ecad2d9d215/config/helm/aws-node-termination-handler/templates/_helpers.tpl#L68-L82

brycahta:
Gotcha, appreciate the explanation. I just merged #604 so closing this PR. Will resolve the issue once released

PR #612, opened 2022-04-04
NTHv2 core functionality

cjerad:
**Issue #, if available:**

**Description of changes:**

* Add SQS parameters to Terminator spec
* Add drain parameters to Terminator spec
* Fetch messages from SQS queue
* Cordon and drain nodes targeted in SQS messages
* Add test suite


src/charts/aws-node-termination-handler-2/templates/node.k8s.aws_terminators.yaml
```diff
@@ -0,0 +1,209 @@
...
+            type: object
+          spec:
+            description: TerminatorSpec defines the desired state of Terminator
+            type: object
+            properties:
+              sqs:
+                description: AWS SQS queue configuration.
+                type: object
+                properties:
+                  attributeNames:
```
bwagner5, line 43: why is this being exposed? Seems like this should be set somewhere else since NTH is what's processing the messages so setting these doesn't really do anything unless NTH knows how to deal with them.

src/charts/aws-node-termination-handler-2/templates/node.k8s.aws_terminators.yaml
```diff
@@ -0,0 +1,209 @@
...
+                  visibilityTimeoutSeconds:
+                    description: |
+                      The duration (in seconds) that the received messages are hidden from subsequent
+                      retrieve requests after being retrieved by a ReceiveMessage request.
+                    type: integer
+                    format: int64
+                    {{- with .Values.terminator.defaults.sqs.visibilityTimeoutSeconds }}
+                    default: {{ . }}
+                    {{- end }}
+                  waitTimeSeconds:
```
bwagner5, line 136: this also seems unnecessary to expose especially since you need knowledge of the http response timeout configured for aws sdk.

src/cmd/controller/main.go
```diff
@@ -55,63 +89,190 @@ func init() {
...
+	config := aws.NewConfig().WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint)
+	sess, err := session.NewSessionWithOptions(session.Options{
+		Config:            *config,
+		SharedConfigState: session.SharedConfigEnable,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create AWS session: %w", err)
+	}
+
+	region, err := ec2metadata.New(sess).Region()
```
bwagner5, line 257: Seems like this is requiring IMDS to be available which won't always be the case. It should only try IMDS as a last resort after the SDK has tried to find the region in all the well known places.

src/cmd/controller/main.go
```diff
@@ -55,63 +89,190 @@ func init() {
...
+	if err != nil {
+		logger.With("error", err).Fatal("failed to create terminator cordon/drain client")
+	}
+
+	rec := terminator.Reconciler{
+		Name:                 "terminator",
+		RequeueInterval:      time.Duration(10) * time.Second,
+		NodeGetter:           nodeGetter,
+		NodeNameGetter:       nodeNameGetter,
+		SqsClientBuilder:     terminatorSqsClientBuilder,
```
bwagner5, line 208: in general initialisms and acronyms should have consistent casing: https://github.com/golang/go/wiki/CodeReviewComments#initialisms 

For some reason go made a distinction between human written code (which should use consistent casing) and machine generated code (which doesn't have to use consistent casing).

bwagner5 (changes requested):

src/charts/aws-node-termination-handler-2/templates/node.k8s.aws_terminators.yaml
cjerad: Placing the parameters in the Terminator allow varying parameters for individual queues, and it consolidates the configuration to a single location.  The values are included in the request ([here](https://github.com/aws/aws-node-termination-handler/pull/612/files#diff-6e7d2748740e5e4f87c2b71eeea3d540f23cd38cc39a609354057aaead18b4ebR59-R76)).

Only the QueueURL field is required when creating a Terminator.  All other fields will be assigned a default value if omitted.  Each default value may be configured via the helm chart ([here](https://github.com/aws/aws-node-termination-handler/pull/612/files#diff-d69c26c948786815eaf97e0cc35a7e4fbac48df08ed71dc0d0f7f65420b7ae92R13-R22)).

src/charts/aws-node-termination-handler-2/templates/node.k8s.aws_terminators.yaml
cjerad: I don't follow what you mean.

As I understand the [docs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html), each request may specify a visibility timeout.  If not specified then a default value from the client or the queue's config will be used. NTH v1 hardcodes the value [here](https://github.com/aws/aws-node-termination-handler/blob/33c6238206b28e38062692188082aaba7287aa0b/pkg/monitor/sqsevent/sqs-monitor.go#L232).

bwagner5: I was referring to `waitTimeSeconds`

cjerad: Thanks for clarifying.  So the value should just be hardcoded like in v1 ([here](https://github.com/aws/aws-node-termination-handler/blob/33c6238206b28e38062692188082aaba7287aa0b/pkg/monitor/sqsevent/sqs-monitor.go#L233))?  For reference, is there documentation why that value was chosen to work with the aws sdk in NTH v1?

bwagner5: it's the max time you can wait on a long poll: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/working-with-messages.html#:~:text=The%20maximum%20long%20polling%20wait%20time%20is%2020%20seconds

src/charts/aws-node-termination-handler-2/templates/node.k8s.aws_terminators.yaml
bwagner5: I realize QueueURL is the only required field, but I'm curious if a user would ever need to configure `attributeNames`. It seems to me that they wouldn't. Having the configuration all in one spot is nice for code neatness, but when thinking about TerminatorSpec as the NTH API for the user, it seems odd to have things you can configure that don't have an affect. 

The attributeNames could still be runtime defaulted without being present in the Spec, and if it's interesting what those values are, you could place them in the `Status` of the Terminator to show what runtime defaults were applied without allowing the user to change them.

src/test/reconciliation_test.go
```diff
@@ -0,0 +1,1666 @@
...
+		completeASGLifecycleActionFunc CompleteASGLifecycleActionFunc
+		describeEC2InstancesFunc       DescribeEC2InstancesFunc
+		kubeGetFunc                    KubeGetFunc
+		receiveSQSMessageFunc          ReceiveSQSMessageFunc
+		deleteSQSMessageFunc           DeleteSQSMessageFunc
+		cordonFunc                     kubectlcordondrain.CordonFunc
+		drainFunc                      kubectlcordondrain.DrainFunc
+	)
+
+	When("the SQS queue is empty", func() {
```
snay2, line 147: [no action needed] Love these tests; they're very easy to reason about.

snay2 (changes requested):

brycahta (commented):
I feel like the preemptive versioning in the `event` pkg (v0, v1, v2) is unnecessary and makes the project structure more confusing. My understanding is if we didn't include these versions now, then when/**IF** we ever release a v2 then we would need to spend additional time to update module versions (potentially logic). Assuming this was the intent, then I don't think the cost is worth the _potential_ savings.

I think it'd be clearer to remove versions altogether. Ex: `pkg/event/rebalancerecommendation/event.go` 

Lmk if this assumption is wrong or if I'm missing context

src/cmd/controller/main.go
```diff
@@ -144,71 +147,32 @@ func main() {
...
-	if err != nil {
-		logger.With("error", err).Fatal("failed to create terminator SQS message client builder")
-	}
-
-	cordonDrainerBuilder, err := kubectlcordondrainer.NewBuilder(
-		clientSet,
-		kubectlcordondrainer.DefaultCordoner,
-		kubectlcordondrainer.DefaultDrainer,
+	eventParser := event.NewAggregatedParser(
+		asgterminateeventv1.Parser{ASGLifecycleActionCompleter: asgClient},
```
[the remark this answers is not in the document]
cjerad: It corresponds to the value of the `"version"` field from the EventBridge message (stored in `AWSMetadata.Version`).  Each event type has a versioned schema defined in EventBridge, e.g. [here](https://github.com/aws/aws-node-termination-handler/pull/612/files#diff-544ce1b445bf7f98041c047ae1a9ce60896914a1c1d80f772bed0f4a8e60a378R26-R27).

src/pkg/event/asgterminate/v2/awsevent.go
```diff
@@ -23,36 +23,36 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
-// AwsEvent contains the properties defined in AWS EventBridge schema
+// AWSEvent contains the properties defined in AWS EventBridge schema
```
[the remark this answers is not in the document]
cjerad: I chose not to use that approach because it spreads the logic needed to support an event across multiple packages, which then may need to be updated if new event types are added or removed.

This approach places all of the logic specific to an event type in a single, dedicated package. Adding or removing support for a new event type is accomplished by simply adding or removing its parser to/from the aggregated list.

The rest of the app interacts with these events only through the [`Event` interface](https://github.com/aws/aws-node-termination-handler/pull/612/files#diff-03278e672080d5c6852e33bfdee29f67f775d3394fbba5636e64dda0627b1eeaR49-R54), so there are no event type checks to handle special-cases for certain events.

src/pkg/event/rebalancerecommendation/v0/awsevent.go
```diff
@@ -23,25 +23,25 @@ import (
 	"go.uber.org/zap/zapcore"
```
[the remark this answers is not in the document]
cjerad: My preference is for smaller, more focused files (e.g. `awsevent.go` = types/logic for unmarshaling, `event.go` = types/logic for handling the event).  Does that shed any light on it? Maybe the files should be renamed? Or I can combine them, if you want.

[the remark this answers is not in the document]
brycahta: Thanks for the context. My vote is still to consolidate into `event.go`. In addition to being closely related, I also think it's clearer for navigating code (such as on GitHub) and the `awsevent` naming is a bit redundant. Either way-- it's a preference so I'm open to more discussion/feel free to open it up to the team for discussion

src/cmd/controller/main.go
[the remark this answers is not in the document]
brycahta: Ahh gotcha, I feel like I'm getting lost in the versions here: CRD/API version, app version, module/pkg version, and now EventBridge schema version.

With that said, I'm still leaning towards removing the `version` (v0, v1) from the project directories/file paths. I don't think the schema version needs to be exposed in the project, but instead handled within an event's `Parse` logic. For example with `asgterminate` event, `Parse` would try to unmarshal into `v1` struct first.. if that fails try `v2`. 
  * Another idea (related to my other comment) is a "super" struct `AsgTerminate` that is a superset of all fields, unmarshal into that, and then check field(s) for version.

Last concern with breaking out by schema version is if `v3` is released tomorrow with very minimal changes, then we need to create a new directory and set of files that would be mostly copy/paste which I don't think is necessary.

src/pkg/event/rebalancerecommendation/v0/awsevent.go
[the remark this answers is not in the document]
snay2: I can see the reasoning for keeping the marshalling logic separate from the event handling logic, and I'm okay with it either way. My main concern is that this structure and division of responsibilities be clear when we go to implement new events in the future.

`awsevent` and `event` seem to me a good enough pattern to follow. But if we're entertaining new names, would it make sense to give them more function-driven names that include words like "marshaller" or "handler"?

src/cmd/controller/main.go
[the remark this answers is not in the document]
cjerad: > if v3 is released tomorrow *with very minimal changes* [emphasis mine]

It is also possible that new versions would include new and important data fields, or require new API calls.

I think the question is: what should NTH do if it encounters a newer version number of a known event type?

1. Ignore it
    * This is the currently implemented behavior; an error is logged
1. Handle it the same as the previous version
    * Should probably log a warning about the unknown version in case there is a failure later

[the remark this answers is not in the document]
brycahta: > I think the question is: what should NTH do if it encounters a newer version number of a known event type?

> 1. Ignore it
> This is the currently implemented behavior; an error is logged
> 2. Handle it the same as the previous version
Should probably log a warning about the unknown version in case there is a failure later

I'd go with number 2 and make it a best effort to handle it (or maybe configurable?). However, I think regardless of how we implement unknown version handling we don't need to expose the version at the package level.

snay2 (approved):

bwagner5 (approved):
lgtm

PR #615, opened 2022-04-08
Fix Helm chart service labels

stevehipwell:
**Description of changes:**

This PR undoes the change made in #604 and fixes the actual issue I introduced in the chart refactor work.

### DIFF
--- a/config/helm/aws-node-termination-handler/templates/_helpers.tpl
+++ b/config/helm/aws-node-termination-handler/templates/_helpers.tpl
@@ -48,13 +48,28 @@ Common labels
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
-app.kubernetes.io/component: {{ .Release.Name }}
 app.kubernetes.io/part-of: {{ .Release.Name }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 helm.sh/chart: {{ include "aws-node-termination-handler.chart" . }}
 {{- with .Values.customLabels }}
 {{ toYaml . }}
 {{- end }}
+{{- end -}}
+
+{{/*
+Deployment labels
+*/}}
+{{- define "aws-node-termination-handler.labelsDeployment" -}}
+{{ include "aws-node-termination-handler.labels" . }}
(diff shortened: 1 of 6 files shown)


config/helm/aws-node-termination-handler/templates/servicemonitor.yaml
```diff
@@ -25,5 +25,5 @@ spec:
   {{- end }}
   selector:
     matchLabels:
-      {{- include "aws-node-termination-handler.labels" . | nindent 6 }}
+      {{- include "aws-node-termination-handler.selectorLabelsDeployment" . | nindent 6 }}
```
snay2, line 28: This is defined here, for reference: https://github.com/aws/aws-node-termination-handler/blob/1c61c2ddc5159350fd8d82ba812291f542a98232/config/helm/aws-node-termination-handler/templates/_helpers.tpl#L71-L74

The important point is that it defines `component` to be `deployment`.

config/helm/aws-node-termination-handler/templates/_helpers.tpl
```diff
@@ -48,7 +48,6 @@ Common labels
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
-app.kubernetes.io/component: {{ .Release.Name }}
```
snay2, line 51: Just to clarify that my understanding of this change: the bug previously was that `component` was set to `aws-node-termination-handler`, which is not correct. The value should have been `daemonset` or `deployment` (depending on the way NTH runs in the cluster). Your solution is to remove this from the generic `labels` and instead assign it in the supersets of `labels` called `labelsDeployment` and `labelsDaemonset`, and then to use the unrelated `selectorLabelsDeployment` for the service monitor template, which separately sets `component` to `deployment`.

Do I understand correctly?

stevehipwell: Yes that is correct. When I refactored the chart I accidentally left the component in the main labels function and forgot to add the component labels functions. Without these functions the component was incorrectly set in the resource labels but the only bug would be the ServiceMonitor as that's the only selector working on these labels.

Issue #619, opened 2022-04-12
TEST_NOTIFICATION event sent to newly added ASG lifecycle hook w/o Event Bridge triggers NTH panic


ngdah:
**Describe the bug**
NTH panics when trying to read from an SQS associated with a newly added ASG lifecycle hook w/o event bridge.

According to the [ASG lifecycle hook documentation](https://docs.aws.amazon.com/autoscaling/ec2/userguide/prepare-for-lifecycle-notifications.html?icmpid=docs_ec2as_console#test-notification-message-example):
> When you first add a lifecycle hook, a test notification message is published to the notification target

This event will be a `autoscaling:TEST_NOTIFICATION` event, which is not supported by NTH. If there are no ASG terminate event sent, NTH will panic and restart.

**Steps to reproduce**
1. Create a new ASG for testing;
2. Create a new ASG lifecycle hook in the new ASG, providing role ARN and SQS ARN to send message to the queue directly;
3. Setup NTH to read form the SQS associated with the lifecycle hook.

**Expected outcome**
NTH should not panic because of `TEST_NOTIFICATION` event

**Application Logs**
```
2022/04/12 00:44:49 INF only lifecycle termination events from ASG to SQS are supported outside EventBridge
[...]

PR #621, opened 2022-04-14
Ignore test notification events

cjerad:
**Issue #, if available:**

#619 TEST_NOTIFICATION event sent to newly added ASG lifecycle hook w/o Event Bridge triggers NTH panic

**Description of changes:**

Added check for TEST_NOTIFICATION lifecycle transistion and logic to skip remaining processing rather than erroring.

### DIFF
--- a/pkg/monitor/sqsevent/sqs-monitor.go
+++ b/pkg/monitor/sqsevent/sqs-monitor.go
@@ -59,6 +59,18 @@ type InterruptionEventWrapper struct {
 	Err               error
 }
 
+type skip struct {
+	err error
+}
+
+func (s skip) Error() string {
+	return s.err.Error()
+}
+
+func (s skip) Unwrap() error {
+	return s.err
+}
+
 // Kind denotes the kind of event that is processed
 func (m SQSMonitor) Kind() string {
 	return SQSTerminateKind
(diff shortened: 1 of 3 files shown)


pkg/monitor/sqsevent/sqs-monitor_test.go
```diff
@@ -144,6 +170,43 @@ func TestMonitor_EventBridgeSuccess(t *testing.T) {
...
+	ec2Mock := h.MockedEC2{
+		DescribeInstancesResp: getDescribeInstancesResp(dnsNodeName, true, true),
+	}
+	drainChan := make(chan monitor.InterruptionEvent, 1)
+
+	sqsMonitor := sqsevent.SQSMonitor{
+		SQS:              sqsMock,
+		EC2:              ec2Mock,
+		ManagedAsgTag:    "aws-node-termination-handler/managed",
+		ASG:              mockIsManagedTrue(nil), // TODO: is this the right mock factory function?
```
snay2, line 193: Has this TODO been resolved?

pkg/monitor/sqsevent/sqs-monitor.go
```diff
@@ -176,6 +203,10 @@ func (m SQSMonitor) processInterruptionEvents(interruptionEventWrappers []Interr
 			log.Warn().Err(eventWrapper.Err).Msg("dropping interruption event for an already terminated node")
```
snay2, line 203: This looks like a slightly different way of handling messages we want to skip (direct reference to a particular error object `ErrNodeStateNotRunning`). Your new solution of the `skip` struct seems more extensible if we encounter similar events in the future. Should we (perhaps in a future PR) refactor `ErrNodeStateNotRunning` to use the new `skip` struct?

pkg/monitor/sqsevent/sqs-monitor_test.go
cjerad: Yes, I just forgot to remove the comment.

snay2 (approved):

PR #624, opened 2022-04-17
Update deployment.yaml file

GastonKanze:
Fixing properties to take the values from the values.yaml file and not to be hardcoded.

**Issue #, if available:** Values were hardcoded and when we are running helm upgrade command and sending the properties with --set command, is not setting the values. 

**Description of changes:**

To test it, you can run the code in the documentation https://github.com/aws/aws-node-termination-handler:

helm upgrade --install aws-node-termination-handler \
  --namespace kube-system \
  --set enableSpotInterruptionDraining="true" \
  --set enableRebalanceMonitoring="true" \
  --set enableScheduledEventDraining="false" \
  
  and you will see that the properties enableRebalanceMonitoring and enableSpotInterruptionDraining are not set. 

### DIFF
--- a/config/helm/aws-node-termination-handler/templates/deployment.yaml
+++ b/config/helm/aws-node-termination-handler/templates/deployment.yaml
@@ -139,13 +139,13 @@ spec:
               value: {{ .Values.webhookTemplate | quote }}
             {{- end }}
             - name: ENABLE_SPOT_INTERRUPTION_DRAINING
-              value: "false"
+              value: {{ .Values.enableSpotInterruptionDraining | quote }}
             - name: ENABLE_SCHEDULED_EVENT_DRAINING
-              value: "false"
+              value: {{ .Values.enableScheduledEventDraining | quote }}
             - name: ENABLE_REBALANCE_MONITORING
-              value: "false"
+              value: {{ .Values.enableRebalanceMonitoring | quote }}
             - name: ENABLE_REBALANCE_DRAINING
-              value: "false"
+              value: {{ .Values.enableRebalanceDraining | quote }}
             - name: ENABLE_SQS_TERMINATION_DRAINING
               value: "true"
             {{- with .Values.awsRegion }}


config/helm/aws-node-termination-handler/templates/deployment.yaml
```diff
@@ -139,13 +139,13 @@ spec:
...
-              value: "false"
+              value: {{ .Values.enableScheduledEventDraining | quote }}
             - name: ENABLE_REBALANCE_MONITORING
-              value: "false"
+              value: {{ .Values.enableRebalanceMonitoring | quote }}
             - name: ENABLE_REBALANCE_DRAINING
-              value: "false"
+              value: {{ .Values.enableRebalanceDraining | quote }}
             - name: ENABLE_SQS_TERMINATION_DRAINING
               value: "true"
```
snay2, line 150: Thanks for fixing this! 😅

Just to confirm, there are two remaining hardcoded booleans in this file:
* `ENABLE_SQS_TERMINATION_DRAINING` is hardcoded `true` because of [line 1](https://github.com/aws/aws-node-termination-handler/blob/ec727aedad3a096f4679f641af9c834c7f840c18/config/helm/aws-node-termination-handler/templates/deployment.yaml#L1) of this file
* `ENABLE_PROBES_SERVER` is hardcoded `true` (line 72) because we expect the probes server to run when NTH is a deployment.

Is that your understanding as well?

GastonKanze: Hi @snay2  , correct, just to add more details:

About **ENABLE_PROBES_SERVER** , in the values.yaml file you can see:

# If the probes server is running for the Daemonset
enableProbesServer: false

for this reason in the daemonset is using the property and not the hardcoded value: https://github.com/aws/aws-node-termination-handler/blob/main/config/helm/aws-node-termination-handler/templates/daemonset.linux.yaml

snay2 (approved):

Issue #494, opened 2021-10-07
Frequent calls to DescribeInstances cause SQS backup and late notifications


mechanical-fish:
`aws-node-termination-handler` seems to have the following logic (in `pkg/monitor/sqsevent/sqs-monitor.go`):

- Pull five messages from the SQS queue every 2 seconds with a 20-second-per-message visibility timeout
- Attempt to call the `ec2.DescribeInstances` API once per message (as part of the `retrieveNodeName` function). If the SQS queue is consistently full, this is 2.5 calls to `DescribeInstances` per second per queue.
- If `DescribeInstances` hits a rate limit, do not remove the message from the queue; instead leave it there to be retried 20 seconds later

We are now operating at sufficient scale that `DescribeInstances` is often hitting API rate limits -- I'm not sure those limits are published, so I'm not sure whether the 2.5-calls-per-second from each `aws-node-termination-handler` are a signifiant cause of our rate limiting, or merely one contributor. But once rate limiting begins we observe the following failure mode:

- `aws-node-termination-handler` will never pull more than 5 messages every 2 seconds.
- If it hits a rate limit the message stays on the queue to be redelivered 20 seconds later.
- The queue gets backed up. We consistently observe 40-50 messages in flight on our busiest SQS queues, though never more than 50 (note that 5 messages * 20 seconds / 2 seconds = 50 messages). There is often a backup of hundreds of messages and the wait times are 50 seconds to 300 seconds or more.
- At this point the majority of messages are handled so late that the instances are gone from EC2. The "good" news is that, when `DescribeInstances` manages to get through to the API, the message for an already-terminated instance gets deleted from the backed-up queue. The bad news is that it's way too late to warn Kubernetes by that point, so we are experiencing sudden unexpected failure of Kube pods.

I can think of a couple ways to change this code to help avoid this problem:

- Stop making so many calls to `DescribeInstances`. The purpose of these calls is to translate EC2 instance IDs to node names. Instance IDs are said to be unique (and are certainly vanishingly unlikely to be reused in rapid succession) and private DNS names are not mutable, so this mapping cannot change and can be cached -- we can periodically call `DescribeInstances` for all instances and cache the results.
- Make the number of SQS messages retrieved per 2-second cycle into a configurable parameter that can be tuned along with the number of workers. I'm not even sure why we fetch so few messages each time, except as a means of throttling the rate of `DescribeInstances` API calls, which could be managed differently as suggested above.

bwagner5:
[...]

PR #625, opened 2022-04-18
add node label selector to Terminator

cjerad:
**Issue #, if available:**

**Description of changes:**

Add `matchLabels` property to the Terminator spec.

NTH will take no action when an SQS message targets a node that does not have the requisite labels and values.

### DIFF
--- a/src/test/reconciliation_test.go
+++ b/src/test/reconciliation_test.go
@@ -229,8 +229,8 @@ var _ = Describe("Reconciliation", func() {
 				Expect(asgLifecycleActions).To(And(HaveKeyWithValue(instanceIDs[1], Equal(StatePending)), HaveLen(1)))
 			})
 
-			It("deletes the message from the SQS queue", func() {
-				Expect(sqsQueues[queueURL]).To(BeEmpty())
+			It("does not delete the message from the SQS queue", func() {
+				Expect(sqsQueues[queueURL]).To(HaveLen(1))
 			})
 		})
 	})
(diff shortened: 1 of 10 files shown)


src/test/reconciliation_test.go
```diff
@@ -1227,6 +1243,91 @@ var _ = Describe("Reconciliation", func() {
...
+			It("returns success and requeues the request with the reconciler's configured interval", func() {
+				Expect(result, err).To(HaveField("RequeueAfter", Equal(reconciler.RequeueInterval)))
+			})
+
+			It("does not cordon or drain any nodes", func() {
+				Expect(cordonedNodes).To(BeEmpty())
+				Expect(drainedNodes).To(BeEmpty())
+			})
+
+			It("does not delete the message from the SQS queue", func() {
```
snay2, line 1325: How long do messages stay on the SQS queue? It seems like there could be a failure mode where NTH spends all its time re-processing messages on the queue (because, e.g., the labels didn't match) and isn't able to respond to new messages in time for k8s to gracefully handle a node termination. That's probably uncommon, but it is basically what mechanical-fish described in https://github.com/aws/aws-node-termination-handler/issues/494.

Is that a concern with this way of handling node labels? It seems like the philosophy of the terminator is "I only respond to requests that match my parameters; anything else I will leave for someone else to handle". So it may not be possible to prevent that situation. Curious what your thoughts are.

cjerad: Queues have a [message retention period](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-configure-queue-parameters.html) setting. A [dead-letter queue](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html) can also be configured.

Leaving the message in the queue gives another Terminator instance a chance to match and handle it.

snay2 (approved):

PR #628, opened 2022-04-20
add event action config to Terminator

cjerad:
**Issue #, if available:**

**Description of changes:**

Add `events` property to the Terminator spec which gives the option to instruct NTH to take no action, or only cordon nodes, when particular message types are received.

### DIFF
--- a/src/pkg/terminator/reconciler.go
+++ b/src/pkg/terminator/reconciler.go
@@ -124,81 +124,80 @@ func (r Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (recon
 		return reconcile.Result{}, err
 	}
 
-	origCtx := ctx
 	for _, msg := range sqsMessages {
-		ctx = logging.WithLogger(origCtx, logging.FromContext(origCtx).
-			With("sqsMessage", logging.NewMessageMarshaler(msg)),
-		)
-
-		evt := r.Parse(ctx, msg)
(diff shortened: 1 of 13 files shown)


src/pkg/terminator/reconciler.go
```diff
@@ -132,9 +133,15 @@ func (r Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (recon
...
 		ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("event", evt))
 
+		evtAction := actionForEvent(evt, terminator)
+		ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("action", evtAction))
+
 		allInstancesHandled := true
+		ec2InstanceIDs := evt.EC2InstanceIDs()
 		savedCtx := ctx
-		for _, ec2InstanceID := range evt.EC2InstanceIDs() {
+		for i := 0; i < len(ec2InstanceIDs) && evtAction != v1alpha1.Actions.NoAction; i++ {
```
snay2, line 142: The condition on this for loop is a little hard to parse. I think the intent is "if `evtAction` is `NoAction`, skip the entire for loop". If that's the case, it would be easier to read if that check were in an explicit `if` statement before the loop starts (but that does increase the nesting on this already heavily nested portion of code). Alternatively, perhaps this is an opportunity to factor out some of the logic into a separate function.

I don't have a strong opinion on this, but I wanted to hear your reasoning behind this choice.

snay2 (approved):

Issue #550, opened 2021-12-17
Queue Processor: Wrong AWS Health Events


joshua-giumelli-deltatre:
In the docs for the infrastructure required for Queue Processor mode, it mentions to create an EventBridge rule for catching `aws.health` messages. The issue with this rule however is that it will catch all AWS health events, even those that cannot be processed by the NTH.

We recently found that this rule was catching notifications for ElasticSearch/OpenSearch relating to the Log4Shell CVE. This was causing the NTH to try and process events for the "ES" service and then would get stuck in a crash loop trying to process this message on the queue.

To avoid this in future, an extra event pattern such as "service": "ec2" would be needed for the `ScheduledChange` rule.

**Steps to reproduce**
1. Create the EventBridge rules from docs for Infrastructure Setup when using Queue Processor
2. Receive non EC2 AWS Health event notification, e.g. for ElasticSearch/OpenSearch Log4J CVE

**Expected outcome**
NTH should either ignore messages for services it is not able to process OR event notifications not relevant to NTH should not end up in its SQS queue.

**Application Logs**
NTH Logs:

```
2021/12/17 03:28:02 ERR ignoring interruption event due to error error="events from Amazon EventBridge for service (ES) are not supported"
2021/12/17 03:28:02 ERR error processing interruption events error="some interruption events for message Id 1100000000000000001000000001100000101000 could not be processed"
2021/12/17 03:28:02 WRN There was a problem monitoring for events error="none of the waiting queue events could be processed" event_type=SQS_TERMINATE
2021/12/17 03:28:02 WRN Stopping NTH - Duplicate Error Threshold hit.
panic: none of the waiting queue events could be processed

goroutine 97 [running]:
main.main.func3(0x1, 0x24ef720, 0xc00007f4a0, 0x21be2d2, 0x1c, 0x0, 0x0, 0x24ff9c8, 0xc000714fc0, 0x24ff9c8, ...)
	/node-termination-handler/cmd/node-termination-handler.go:211 +0x649
created by main.main
	/node-termination-handler/cmd/node-termination-handler.go:193 +0xc91
```

Problem Event Notification:

```
{
    "version": "0",
    "id": "9ca40099-e97d-510f-e9df-6f92042f34e1",
    "detail-type": "AWS Health Event",
    "source": "aws.health",
    "account": "XXXXXX",
    "time": "2021-12-14T04:05:00Z",
    "region": "ap-south-1",
    "resources": [
        "XXXXXX"
    ],
[...]

PR #633, opened 2022-04-26
Fix AWS Health Event Bridge Rule

LikithaVemulapalli:
**Issue #, if available:** https://github.com/aws/aws-node-termination-handler/issues/550

**Description of changes:**  
- Modified the Event Bridge Rule for "MyK8sScheduledChangeRule" in README document to consider the detail-type for the AWS Health Events.
- Modified the conditions in "scheduled-change-event" file for checking service and event type category for AWS Health Events. Instead of returning an error, now it returns a warning message and deletes the messages that are not handled by NTH.

- For testing the changes, followed the below steps:
                       - Ran the e2e tests to check whether the modified changes are affecting any  other functionalities.
                       - Tried to send an AWS health event from different service other than EC2 from SQS service.
                       - Verified if the sent message has been deleted and a warning message is returned.
                       - Tried to send an AWS Health Event with EC2 service and verified if the output for that Event is as expected.

### DIFF
--- a/pkg/monitor/sqsevent/scheduled-change-event.go
+++ b/pkg/monitor/sqsevent/scheduled-change-event.go
@@ -76,13 +76,13 @@ func (m SQSMonitor) scheduledEventToInterruptionEvents(event *EventBridgeEvent, 
 	}
 
 	if scheduledChangeEventDetail.Service != "EC2" {
-		log.Warn().Msgf("events from Amazon EventBridge for service (%s) are not supported", scheduledChangeEventDetail.Service)
-		return append(interruptionEventWrappers, InterruptionEventWrapper{nil, nil})
+		err := skip{fmt.Errorf("events from Amazon EventBridge for service (%s) are not supported", scheduledChangeEventDetail.Service)}
+		return append(interruptionEventWrappers, InterruptionEventWrapper{nil, err})
 	}
 
 	if scheduledChangeEventDetail.EventTypeCategory != "scheduledChange" {
-		log.Warn().Msgf("events from Amazon EventBridge with EventTypeCategory (%s) are not supported", scheduledChangeEventDetail.EventTypeCategory)
-		return append(interruptionEventWrappers, InterruptionEventWrapper{nil, nil})
+		err := skip{fmt.Errorf("events from Amazon EventBridge with EventTypeCategory (%s) are not supported", scheduledChangeEventDetail.EventTypeCategory)}
+		return append(interruptionEventWrappers, InterruptionEventWrapper{nil, err})
 	}
 
 	for _, affectedEntity := range scheduledChangeEventDetail.AffectedEntities {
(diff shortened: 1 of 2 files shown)


pkg/monitor/sqsevent/scheduled-change-event.go
```diff
@@ -76,13 +76,13 @@ func (m SQSMonitor) scheduledEventToInterruptionEvents(event *EventBridgeEvent,
 	}
 
 	if scheduledChangeEventDetail.Service != "EC2" {
-		err := fmt.Errorf("events from Amazon EventBridge for service (%s) are not supported", scheduledChangeEventDetail.Service)
-		return append(interruptionEventWrappers, InterruptionEventWrapper{nil, err})
```
AustinSiu, line 80: This error indicates something that requires action from the user, to update their configuration, rather than just providing a heads-up of an application oddity. With that perspective, I think keeping it as a (skippable) error makes more sense (I know I mentioned this idea in the first place! Sorry!)

It looks like we've recently added support for a skippable type of error: https://github.com/aws/aws-node-termination-handler/blob/37e989956e163b027cd2dcc04a58ce89a75244c6/pkg/monitor/sqsevent/sqs-monitor.go#L58 Switching to use this should be a pretty similar change 🙂

pkg/monitor/sqsevent/scheduled-change-event.go
```diff
@@ -76,13 +76,13 @@ func (m SQSMonitor) scheduledEventToInterruptionEvents(event *EventBridgeEvent,
...
+		log.Warn().Msgf("events from Amazon EventBridge for service (%s) are not supported", scheduledChangeEventDetail.Service)
+		return append(interruptionEventWrappers, InterruptionEventWrapper{nil, nil})
 	}
 
 	if scheduledChangeEventDetail.EventTypeCategory != "scheduledChange" {
-		err := fmt.Errorf("events from Amazon EventBridge with EventTypeCategory (%s) are not supported", scheduledChangeEventDetail.EventTypeCategory)
-		return append(interruptionEventWrappers, InterruptionEventWrapper{nil, err})
```
AustinSiu, line 85: Same thoughts here!

AustinSiu (commented):
Good work! Just a few comments.

AustinSiu (approved):

vincentgna:
this PR missed the cfn sample for this change:

> - Modified the Event Bridge Rule for "MyK8sScheduledChangeRule" in README document to consider the detail-type for the AWS Health Events.

https://github.com/aws/aws-node-termination-handler/blob/82f3a250096706c002eb93527cd99b5fb18f621d/README.md?plain=1#L340

vs

https://github.com/aws/aws-node-termination-handler/blob/82f3a250096706c002eb93527cd99b5fb18f621d/docs/cfn-template.yaml#L35-L42

LikithaVemulapalli:
Thank you @vincentgna for noticing it, will update cfn sample as well.

PR #641, opened 2022-05-18
add webhook config to Terminator

cjerad:
**Issue #, if available:**

**Description of changes:**

Add `webhook` property to `Terminator` spec

### DIFF
--- a/src/pkg/terminator/eventkind.go
+++ b/src/pkg/terminator/eventkind.go
@@ -16,7 +16,7 @@ limitations under the License.
 
 package terminator
 
-type EventKind string
+type EventKind = string
 
 var EventKinds = struct {
 	AutoScalingTermination,
@@ -32,8 +32,4 @@ var EventKinds = struct {
 	SpotInterruption:        EventKind("spotInterruption"),
 	StateChange:             EventKind("stateChange"),
 	Noop:                    EventKind("noop"),
-}
-
-func (e EventKind) String() string {
-	return string(e)
 }
--- a/src/pkg/webhook/builder.go
+++ b/src/pkg/webhook/builder.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2022 Amazon.com, Inc. or its affiliates. All rights reserved.
+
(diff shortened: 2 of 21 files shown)


src/pkg/terminator/eventkind.go
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 
 package terminator
 
-type EventKind string
+type EventKind = string
```
snay2, line 19: Can you describe the semantic difference between these two ways of describing the type? I'm not very familiar with it.

src/pkg/webhook/builder.go
```diff
@@ -0,0 +1,79 @@
...
+
+type (
+	HttpSendFunc = func(*http.Request) (*http.Response, error)
+	ProxyFunc    = func(*http.Request) (*url.URL, error)
+
+	ClientBuilder func(ProxyFunc) HttpSendFunc
+)
+
+func NewHttpClientDo(proxy ProxyFunc) HttpSendFunc {
+	c := &http.Client{
```
snay2, line 36: Does the HTTP client need any retry or backoff configurations, or is this sufficient for how it makes requests?

snay2 (approved):

src/pkg/terminator/eventkind.go
cjerad:
```go
type EventKind string // Defines a new type whose content is a `string`.
type EventKind = string // Defines an alias for the `string` type, i.e. `EventKind` and `string` values can be used interchangeably.
```

The switch made things a easier when including the event kind description in the webhook notification.

src/pkg/webhook/builder.go
cjerad: This is how NTHv1 configures its client (see [here](https://github.com/aws/aws-node-termination-handler/blob/5e4984f2d8588ce6661cc33ccb836249a85852a8/pkg/webhook/webhook.go#L93-L108)).  Might be worth adding retry, etc. support if we run into problems.

PR #644, opened 2022-05-24
delete test notification

cjerad:
**Issue #, if available:**

#619 TEST_NOTIFICATION event sent to newly added ASG lifecycle hook w/o Event Bridge triggers NTH panic

**Description of changes:**

Previous fix did not delete the message from the SQS queue so it would be processed multiple times, each creating an entry in the logs.

Now delete the message when a test notification is detected.


pkg/monitor/sqsevent/sqs-monitor.go
```diff
@@ -138,7 +141,11 @@ func (m SQSMonitor) processLifecycleEventFromASG(message *sqs.Message) (EventBri
 
 	case lifecycleEvent.Event == TEST_NOTIFICATION || lifecycleEvent.LifecycleTransition == TEST_NOTIFICATION:
 		log.Warn().Msg("ignoring ASG test notification")
```
brycahta, line 143: nit: no longer ignoring this notification

brycahta (approved):
/lgtm

PR #645, opened 2022-05-25
Log IMDS response full text when status code not in 200 range

snay2:

**Description of changes:**

Sometimes IMDS may return non-200 status codes. We'd like to see the body of the response in those scenarios, in case there may be useful debugging data there.

Restructured the `GetMetadataInfo()` function to print response body with debug-level logging.

### DIFF
--- a/pkg/ec2metadata/ec2metadata.go
+++ b/pkg/ec2metadata/ec2metadata.go
@@ -195,20 +195,24 @@ func (e *Service) GetRebalanceRecommendationEvent() (rebalanceRec *RebalanceReco
 
 // GetMetadataInfo generic function for retrieving ec2 metadata
 func (e *Service) GetMetadataInfo(path string) (info string, err error) {
+	metadataInfo := ""
 	resp, err := e.Request(path)
-	if err != nil || resp == nil {
+	if err != nil {
 		return "", fmt.Errorf("Unable to parse metadata response: %w", err)
 	}
-	defer resp.Body.Close()
-	body, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return "", fmt.Errorf("Unable to parse http response. Status code: %d. %w", resp.StatusCode, err)
+	if resp != nil {
+		defer resp.Body.Close()
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return "", fmt.Errorf("Unable to parse http response. Status code: %d. %w", resp.StatusCode, err)
+		}
+		metadataInfo = string(body)
+		if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+			log.Info().Msgf("Metadata response status code: %d. Body: %s", resp.StatusCode, metadataInfo)
+			return "", fmt.Errorf("Metadata request received http status code: %d", resp.StatusCode)
+		}
 	}
-	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
-		log.Debug().Msgf("Metadata response status code: %d. Body: %s", resp.StatusCode, string(body))
-		return "", fmt.Errorf("Metadata request received http status code: %d", resp.StatusCode)
-	}
-	return string(body), nil
+	return metadataInfo, nil
 }
 
 // Request sends an http request to IMDSv1 or v2 at the specified path
--- a/pkg/ec2metadata/ec2metadata.go
+++ b/pkg/ec2metadata/ec2metadata.go
@@ -196,16 +196,17 @@ func (e *Service) GetRebalanceRecommendationEvent() (rebalanceRec *RebalanceReco
 // GetMetadataInfo generic function for retrieving ec2 metadata
 func (e *Service) GetMetadataInfo(path string) (info string, err error) {
 	resp, err := e.Request(path)
-	if resp != nil && (resp.StatusCode < 200 || resp.StatusCode >= 300) {
-		return "", fmt.Errorf("Metadata request received http status code: %d", resp.StatusCode)
-	}
-	if err != nil {
+	if err != nil || resp == nil {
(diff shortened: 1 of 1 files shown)


pkg/ec2metadata/ec2metadata.go
```diff
@@ -196,16 +196,17 @@ func (e *Service) GetRebalanceRecommendationEvent() (rebalanceRec *RebalanceReco
 // GetMetadataInfo generic function for retrieving ec2 metadata
 func (e *Service) GetMetadataInfo(path string) (info string, err error) {
 	resp, err := e.Request(path)
-	if resp != nil && (resp.StatusCode < 200 || resp.StatusCode >= 300) {
-		return "", fmt.Errorf("Metadata request received http status code: %d", resp.StatusCode)
-	}
-	if err != nil {
+	if err != nil || resp == nil {
 		return "", fmt.Errorf("Unable to parse metadata response: %w", err)
```
AustinSiu, line 200: nit: In the event of `resp == nil` and `err == nil` both being `true`, this log would look strange. Should we have separate log for "response was empty"?

AustinSiu (dismissed):

pkg/ec2metadata/ec2metadata.go
snay2: Based on the three `return` statements in `Request()`, it's not possible for them to both be nil (i.e., if `resp` is nil, then `err` will describe why). So it's sufficient to check only `err != nil`.

I'll remove the `|| resp == nil` part of the statement to clarify that and keep it consistent with the style of the rest of this function.

snay2: Based on the three `return` statements in `Request()`, it's not possible for them to both be nil (i.e., if `resp` is nil, then `err` will describe why). So it's sufficient to check only `err != nil`.

I'll remove the `|| resp == nil` part of the statement to clarify that and keep it consistent with the style of the rest of this function.

pkg/ec2metadata/ec2metadata.go
```diff
@@ -196,16 +196,17 @@ func (e *Service) GetRebalanceRecommendationEvent() (rebalanceRec *RebalanceReco
...
 	}
 	defer resp.Body.Close()
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return "", fmt.Errorf("Unable to parse http response: %w", err)
+		return "", fmt.Errorf("Unable to parse http response. Status code: %d. %w", resp.StatusCode, err)
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		log.Debug().Msgf("Metadata response status code: %d. Body: %s", resp.StatusCode, string(body))
```
brycahta, line 208: Any thoughts on always printing StatusCode and body for log-level debug and for these non-success status codes always print for log-level info/warning? My thinking is the non-success statuses can impact how NTH runs (i.e. terminate if errThreshold is hit) so that information should be more readily accessible.

snay2: Non-success status codes will already be in the logs, regardless of log level, because they are in the error message.

We don't yet know how useful (or big) the body will be, so I'm more inclined to introduce it at log-level debug so that it won't affect existing customers who don't want or need that extra information. If it proves useful, though, we should definitely log it at info level instead.

Logging success status codes and their bodies is probably a bit superfluous, because we can assume those requests from their downstream effects (i.e., if you're querying for an instance ID and you get it successfully, we will know it was successful because the caller did not fail out and log something).

pkg/ec2metadata/ec2metadata.go
```diff
@@ -196,16 +196,17 @@ func (e *Service) GetRebalanceRecommendationEvent() (rebalanceRec *RebalanceReco
...
 	if err != nil {
 		return "", fmt.Errorf("Unable to parse metadata response: %w", err)
 	}
 	defer resp.Body.Close()
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return "", fmt.Errorf("Unable to parse http response: %w", err)
+		return "", fmt.Errorf("Unable to parse http response. Status code: %d. %w", resp.StatusCode, err)
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```
brycahta, line 207: Can we add back the `resp != nil` check before accessing `StatusCode`? I know `resp` can't be nil if we add the update from my other comment to `Request()`, but not 100% certain without that change..

snay2: If `resp` is nil, we would get an error on line 202 or 203 before even getting here, so I don't think an additional nil check here would be very useful. I'll add it back to line 199 instead so we can catch it before trying to reference any properties.

brycahta: > If resp is nil, we would get an error on line 202 or 203 before even getting here

Can you help me understand where the error gets assigned when resp is nil? Looking thru the code it seems like it's possible to return nil response and no error. Unless the `Request()` method has been refactored recently, I think the previous `!= nil` check is still necessary.

> I'll add it back to line 199 instead so we can catch it before trying to reference any properties.

Idk if I'd say idiomatic, but usually common to see the `!= nil` check coupled with the use (as it was previously).

pkg/ec2metadata/ec2metadata.go
snay2: Discussed offline. We did not previously have a nil check on `resp`, which is probably incorrect. I've added one in the most recent commit.

pkg/ec2metadata/ec2metadata.go
snay2: Discussed offline. The new `resp != nil` check should address this concern.

brycahta (approved):
/lgtm

PR #646, opened 2022-05-26
add make target explore-test-coverage

cjerad:
**Issue #, if available:**

The "grouped targets" feature requires Make 4.3 or newer.

**Description of changes:**

* Added check for Make 4.3 or newer in `Makefile`
* Updated test workflow
* Added shortcut make target `explore-test-coverage`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,6 +33,12 @@ jobs:
           ~/go/bin/
         key: gocache
 
+    - name: Upgrade Make
+      run: |
+        sudo apt-get update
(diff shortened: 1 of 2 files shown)


brycahta (commented):
I'm fine with the `explore-test-coverage` target, but I don't agree with grouped targets and requiring a specific Make version. It makes the Makefile more difficult to read and less explicit, what's the benefit?

pdk27:
Agree with @brycahta. The current version is more readable 👍 .

brycahta (approved):

pdk27 (approved):

PR #657, opened 2022-06-29
Pass `CHECK_ASG_TAG_BEFORE_DRAINING`-variable in daemonsets

atarax:
**Issue #, if available:**

**Description of changes:**
`CHECK_ASG_TAG_BEFORE_DRAINING`-env-var is not passed to daemonset but it is to deployment.

### DIFF
--- a/config/helm/aws-node-termination-handler/templates/daemonset.windows.yaml
+++ b/config/helm/aws-node-termination-handler/templates/daemonset.windows.yaml
@@ -77,6 +77,8 @@ spec:
               value: {{ .Values.probes.httpGet.port | quote }}
             - name: PROBES_SERVER_ENDPOINT
               value: {{ .Values.probes.httpGet.path | quote }}
+            - name: CHECK_ASG_TAG_BEFORE_DRAINING
+              value: {{ .Values.checkASGTagBeforeDraining | quote }}
             - name: LOG_LEVEL
               value: {{ .Values.logLevel | quote }}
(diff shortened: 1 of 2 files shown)


config/helm/aws-node-termination-handler/templates/daemonset.windows.yaml
```diff
@@ -89,6 +89,8 @@ spec:
             - name: INSTANCE_METADATA_URL
               value: {{ . | quote }}
             {{- end }}
+            - name: CHECK_ASG_TAG_BEFORE_DRAINING
+              value: {{ .Values.checkASGTagBeforeDraining | quote }}
```
snay2, line 93: Could you move this up below line 79? Keeping the parameters in the same order will make it easier to compare the linux and windows versions of this file in the future.

snay2 (changes requested):
Can you update the PR summary to describe the testing you did to verify the behavior before/after this change?

config/helm/aws-node-termination-handler/templates/daemonset.windows.yaml
atarax: sure, no big deal :)

bwagner5:
@atarax This is because the `CHECK_ASG_TAG_BEFORE_DRAINING` is only for queue-processor mode of NTH which runs as a deployment. IMDS mode (DaemonSet mode) does not check tags at all before draining.

PR #672, opened 2022-08-15
Update golangci lint action

brycahta:

**Description of changes:** Add `setup-go` step with go version to  `golangci` job to ensure intended go version is used

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -13,7 +13,7 @@ on:
     - cron: '0 19 * * 1-5'
 
 env:
-  DEFAULT_GO_VERSION: ^1.16
+  DEFAULT_GO_VERSION: 1.16
   GITHUB_USERNAME: ${{ secrets.EC2_BOT_GITHUB_USERNAME }}
   GITHUB_TOKEN: ${{ secrets.EC2_BOT_GITHUB_TOKEN }}
   WEBHOOK_URL: ${{ secrets.WEBHOOK_URL }}
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -59,12 +59,6 @@ jobs:
     name: Lint Eastwood
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/setup-go@v2
-        with:
-          go-version: "1.16"
-      - run: |
-          export PATH=${GOROOT}/bin:$PATH
-          go version
       - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -93,10 +93,10 @@ helm-validate-eks-versions:
 	${MAKEFILE_PATH}/test/helm/validate-chart-versions
 
 build-binaries:
-	${MAKEFILE_PATH}/scripts/build-binaries -p ${SUPPORTED_PLATFORMS_LINUX} -v ${VERSION} -d
```
snay2, line 96: The `-d` or `GOPROXY=direct` arguments on the various build commands are important to being able to build locally, when the developer has certain firewall rules in place on their local machine. Is there a way we can keep the no-proxy mode for local builds while discarding it for GitHub Actions builds?

One possible solution is to examine the `CI` environment variable, which [is always set to `true` in GHA](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables). Maybe we can create a makefile variable that contains either `-d` or nothing, depending on whether `CI` is set, and each relevant build command can include that variable.

.github/workflows/build-and-test.yaml
```diff
@@ -13,7 +13,7 @@ on:
     - cron: '0 19 * * 1-5'
 
 env:
-  DEFAULT_GO_VERSION: ^1.16
+  DEFAULT_GO_VERSION: 1.16
```
snay2, line 16: It seems like `^1.16` [meant](https://github.com/npm/node-semver#caret-ranges-123-025-004) "any minor version >= 1.16 but < 2.0" (thus, `1.17` or even `1.19` would be acceptable). Removing the caret means we want exactly `1.16`.

Do I understand your intent correctly?

Makefile
brycahta: I agree with the idea to improve setting `GOPROXY=direct` for *when the developer has certain firewall rules in place on their local machine*. However, I would think these devs are the minority and we wouldn't want to assume all devs can't access the general Go module repo (`https://proxy.golang.org`). Brainstorming some other ideas:
* add helper method, `set_goproxy()` to scripts that need `-d` / `direct` flag. This would ping `https://proxy.golang.org` and if response is not success, then set to `direct`
* add failure messaging to scripts that _suggests_ setting `GOPROXY=direct` based on certain error messaging
* add makefile local targets, ex: `make build-binaries-local`. either sets `GOPROXY=direct` or passes `-d`

brycahta:
Cleaning this up in subsequent PR

PR #485, opened 2021-08-24
Add extraEnv to helm chart

jrsherry:
Issue #, if available:
https://github.com/aws/aws-node-termination-handler/issues/484

Description of changes:
This change will allow for additional environment variables to be passed to the NTH pod (i.e. proxy config).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #674, opened 2022-08-16
helm: Apply extraEnv to daemonsets

hamishforbes:
Looks like this was (accidentally I presume) removed in #540 

I *think* helm changes are made here and not on the eks-charts repo?
If that's not correct I'm happy to open this there instead!

### DIFF
--- a/config/helm/aws-node-termination-handler/templates/daemonset.linux.yaml
+++ b/config/helm/aws-node-termination-handler/templates/daemonset.linux.yaml
@@ -151,6 +151,9 @@ spec:
               value: "false"
             - name: UPTIME_FROM_FILE
               value: {{ .Values.procUptimeFile | quote }}
+            {{- with .Values.extraEnv }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
           {{- if or .Values.enablePrometheusServer .Values.enableProbesServer }}
           ports:
           {{- if .Values.enableProbesServer }}
--- a/config/helm/aws-node-termination-handler/templates/daemonset.windows.yaml
+++ b/config/helm/aws-node-termination-handler/templates/daemonset.windows.yaml
@@ -149,6 +149,9 @@ spec:
               value: {{ .Values.enableRebalanceDraining | quote }}
             - name: ENABLE_SQS_TERMINATION_DRAINING
               value: "false"
+            {{- with .Values.extraEnv }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
           {{- if or .Values.enablePrometheusServer .Values.enableProbesServer }}
           ports:
           {{- if .Values.enableProbesServer }}


config/helm/aws-node-termination-handler/templates/daemonset.linux.yaml
```diff
@@ -151,6 +151,9 @@ spec:
               value: "false"
             - name: UPTIME_FROM_FILE
               value: {{ .Values.procUptimeFile | quote }}
+            {{- with .Values.extraEnv }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
```
snay2, line 156: When this `extraEnv` block was originally added to the daemonset and deployment templates in #485, it broke out the values into explicit `name` and `value` parameters ([source](https://github.com/aws/aws-node-termination-handler/blob/bef4a565c0efcca22ad916cd2d417c2598a5913e/config/helm/aws-node-termination-handler/templates/daemonset.linux.yaml#L179-L182)):

```yaml
{{- range $key, $value := .Values.extraEnv }}
          - name: {{ $key }}
            value: {{ $value | quote }}
{{- end }}
```

However, during the big refactor in #540, that was changed to the format you're using here ([source](https://github.com/aws/aws-node-termination-handler/blob/main/config/helm/aws-node-termination-handler/templates/deployment.yaml#L167-L169)):

```yaml
            {{- with .Values.extraEnv }}
              {{- toYaml . | nindent 12 }}
            {{- end }}
```

I'm assuming this format is sufficient, since it is working well enough for folks who use NTH as a deployment. Can you confirm from your testing that simply embedding with the `toYaml` directive works correctly for a daemonset?

snay2 (commented):
Yes, you are correct that this repo is the right place to make helm chart changes. After we release here, the changes will propagate to the eks-charts repo.

I agree that removing `extraEnv` from the daemonset templates (but leaving it in the deployment template) was an oversight during the refactor in #540. Thanks for fixing it!

config/helm/aws-node-termination-handler/templates/daemonset.linux.yaml
hamishforbes: The original method actually accepted a map of env vars like

```
extraEnv:
  FOOVAR: BARVAL
```

even though the default value in `values.yaml` was an empty array.

Technically #540 should've been a breaking change for `extraEnv` even on the `deployment` as it now accepts

```
extraEnv:
  - name: FOOVAR
    value: BARVAL
```

Personally I prefer the map, its less verbose in the values, but that would be a breaking change from the current behaviour!

It does work with the `toYaml` though:

```
> helm template -s templates/daemonset.linux.yaml --set extraEnv[0].name=FOO --set extraEnv[0].value=BAR nth . | grep FOO -A1 -B2
            - name: UPTIME_FROM_FILE
              value: "/proc/uptime"
            - name: FOO
              value: BAR
```

snay2: You're right, #540 should have been a breaking change for several reasons, including this. I incorrectly tagged it as a patch release (v1.14.1), and after a long back-and-forth (on #565 and #569) we decided to live with the mistake.

Thanks for the explanation here! Glad it's working, even if not ideal.

snay2 (approved):

snay2:
This has been released in `v1.17.1`, chart version `0.19.1`.

PR #669, opened 2022-08-09
Filter managed non-ASG nodes by tag

AustinSiu:
**Issue #, if available:** #654

**Description of changes:**
* Replaced fields `check-asg-tag-before-draining` and `managed-asg-tag` with `check-tag-before-draining` and `managed-tag`, updated docs to mark them as deprecated, and removed them from charts
* Removed ASG api calls to fetch ASG info when determining whether a node should be managed by NTH.
* Fixed bug where the presence of the `managedAsgTag/managedTag` on an instance wouldn't be checked correctly before deferring to ASG lookup

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #654, opened 2022-06-17
filter non-ASG nodes by tag 


universam1:
**Describe the feature**

Karpenter nodes are not bound to a Autoscaling Group. Filter on `tags` via `check-asg-tag-before-draining` is scoped to ASG only, so it is not possible to filter nodes such nodes based on tags.
https://github.com/aws/aws-node-termination-handler/blob/2048e9263d2b039a22cffd2c34b55e7658b52261/pkg/monitor/sqsevent/sqs-monitor.go#L356-L365

**Is the feature request related to a problem?**

AFAIK pre - filtering the `Spot ITN Events` is not possible because `EC2 Spot Instance Interruption Warning` do not contain `tags` of the instance where we could filter the rule on

I have several clusters in one account and thus I am missing an option in NTH to filter for cluster respective nodes only.

**Describe alternatives you've considered**

none

AustinSiu:
Hi @universam1! Can you help confirm the ask? It sounds like you want the ability for NTH to ignore nodes that are not associated with an ASG, using the same `ManagedAsgTag` we use today for nodes in ASGs (name might need to be revisited...).

> I have to bound NTH to the cluster respective nodes only.

Can you help clarify what you mean, assuming this is your current workaround to the issue?

universam1:
@AustinSiu Thank you for asking - actually the opposite 😄 

I want NTH to manage non-ASG nodes that are created by Karpenter. Plus I need to filter based on tags to identify the cluster - nodes. 
However, `ManagedAsgTag` filter is not applicable for non-ASG nodes, see above.

> assuming this is your current workaround to the issue?

Sorry if I was unclear, I have no workaround for this problem.

bwagner5:
👋 @universam1 you are right and this is a regression. We used to have an `AssumeASGTagPropagation` configuration that would result in no ASG lookup, but that was removed recently.  We should add that back. 

https://github.com/aws/aws-node-termination-handler/commit/37e989956e163b027cd2dcc04a58ce89a75244c6#r78210273

pangorgo:
I've just encountered a similar issue. I just wanted to post a potential workaround to tag your instance with `aws:autoscaling:groupName`, but I forgot `aws:` prefix is reserved and the user can't assign such a tag.

[...]

PR #681, opened 2022-08-31
Remove unit test functions that mock ASG managed tag

snay2:
**Issue #, if available:** #654

**Description of changes:**

PR #669 removed calls to ASG APIs when checking for the "managed" tag on EC2 instances. Some of the unit tests were still testing that behavior, which is no longer necessary. This PR removes those functions.

A few of the unit tests do not need to set an `ASG` parameter at all, so I removed those lines entirely. In tests that do need the ASG parameter, I left it but passed in a default ASG mock object.

**Testing:**

Verified that unit tests pass after these changes. No changes made to functional production code, only unit tests.


pdk27 (approved):
LGTM

PR #684, opened 2022-09-06
fix(sqs): log itn event messageid correctly

acobaugh:
**Description of changes:**

This fixes the logging so we log the MessageId, not a binary representation of the pointer. Otherwise we get messages like:

`2022/09/06 19:22:45 ERR error processing interruption events error="some interruption events for message Id 1100000000000000001111000101100011101000 could not be processed"`

### DIFF
--- a/pkg/monitor/sqsevent/sqs-monitor.go
+++ b/pkg/monitor/sqsevent/sqs-monitor.go
@@ -241,7 +241,7 @@ func (m SQSMonitor) processInterruptionEvents(interruptionEventWrappers []Interr
 	}
 
 	if failedInterruptionEventsCount != 0 {
-		return fmt.Errorf("some interruption events for message Id %b could not be processed", message.MessageId)
+		return fmt.Errorf("some interruption events for message Id %s could not be processed", *message.MessageId)
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


pkg/monitor/sqsevent/sqs-monitor.go
```diff
@@ -241,7 +241,7 @@ func (m SQSMonitor) processInterruptionEvents(interruptionEventWrappers []Interr
 	}
 
 	if failedInterruptionEventsCount != 0 {
-		return fmt.Errorf("some interruption events for message Id %b could not be processed", message.MessageId)
+		return fmt.Errorf("some interruption events for message Id %s could not be processed", *message.MessageId)
```
snay2, line 244: Thanks for fixing this!

One of the [unit tests is failing](https://github.com/aws/aws-node-termination-handler/runs/8215609377?check_suite_focus=true#step:5:578) because our mock SQS messages don't have a `MessageId` property in them. Can you add it? Replace the last line of `getSQSMessageFromEvent()` of [`pkg/monitor/sqsevent/sqs-monitor_test.go`](https://github.com/aws/aws-node-termination-handler/blob/31f4b9f2e2f2e1fd35d558f14e4e97673acebee8/pkg/monitor/sqsevent/sqs-monitor_test.go#L915-L922) with something like this:

```
messageId := "1234567890"
return sqs.Message{Body: &eventStr, MessageId: &messageId}, nil
```

snay2 (changes requested):
One of the unit tests is failing with this change

acobaugh:
Fixed.

snay2 (approved):
LGTM!

PR #697, opened 2022-10-03
add sync-to-aws-eks-charts.sh script

cjerad:
**Issue #, if available:**

**Description of changes:**

Added script that creates a PR to aws/eks-charts with the latest NTHv2 helm chart.

### DIFF
--- a/scripts/sync-to-aws-eks-charts.sh
+++ b/scripts/sync-to-aws-eks-charts.sh
@@ -1,0 +1,198 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+repo_root_path="$(cd "$(dirname "$0")"; cd ..; pwd -P)"
+makefile_path="${repo_root_path}/Makefile"
+
+if ! command -v gh >/dev/null; then
+    echo "error: required executable 'gh' not found" >&2
+    exit 1
+fi
+
+github_nth_full_repo_name="$(make -s -f "${makefile_path}" repo-full-name)"
+nth_version="$(make -s -f "${makefile_path}" version)"
+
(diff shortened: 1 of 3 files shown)


scripts/sync-to-aws-eks-charts.sh
```diff
@@ -0,0 +1,198 @@
...
+
+github_nth_full_repo_name="$(make -s -f "${makefile_path}" repo-full-name)"
+nth_version="$(make -s -f "${makefile_path}" version)"
+
+helm_chart_name="aws-node-termination-handler-2"
+helm_chart_path="${repo_root_path}/charts/${helm_chart_name}"
+
+github_eks_charts_full_repo_name="aws/eks-charts"
+github_username="${GITHUB_USERNAME:-}"
+github_token="${GITHUB_TOKEN:-}"
```
snay2, line 21: If I understand correctly (per [these docs](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Parameter-Expansion)), `:-` means that if `GITHUB_USERNAME` is unset or nil, the resulting value in `github_username` will be empty string (since nothing follows the hyphen). Is that right? Is that better than merely using `${GITHUB_USERNAME}`?

cjerad: The final check that both a username and token have been provided is done on lines 77 & 78.  Here the `:-` is just to prevent the script from terminating if either variable is undefined (due to `set -u` on line 3) since the values may be provided via command line options.

snay2: Cool, thanks for the explanation!

snay2 (approved):
lgtm

PR #698, opened 2022-10-05
add sync-readme-to-ecr-public.sh script

cjerad:
**Issue #, if available:**

**Description of changes:**

Add script that uploads the contents of README.md (up to the max allowed length) to the ECR public repos for the controller and webhook.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,9 +15,10 @@ KODATA = \
 	cmd/webhook/kodata/HEAD \
 	cmd/webhook/kodata/refs
 CODECOVERAGE_OUT = $(PROJECT_DIR)/coverprofile.out
-GITHUB_REPO_FULL_NAME = "aws/aws-node-termination-handler"
-ECR_PUBLIC_REGISTRY ?= "public.ecr.aws/aws-ec2"
-ECR_PUBLIC_REPOSITORY_ROOT = "aws-node-termination-handler-2"
+GITHUB_REPO_FULL_NAME = aws/aws-node-termination-handler
+ECR_PUBLIC_REGION = us-east-1
+ECR_PUBLIC_REGISTRY ?= public.ecr.aws/aws-ec2
+ECR_PUBLIC_REPOSITORY_ROOT = aws-node-termination-handler-2
 
 # Image URL to use all building/pushing image targets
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -129,10 +130,14 @@ apply: $(KO) $(KODATA) ## Deploy the controller into the current kubernetes clus
 delete:  ## Delete controller from current kubernetes cluster.
 	helm uninstall dev --namespace ${CLUSTER_NAMESPACE}
 
+.PHONY: ecr-login
+ecr-login: ## Login to default AWS ECR repository.
```
snay2, line 134: This target doesn't appear to be used anywhere. Is it worth keeping (e.g., for testing, etc.)?

scripts/sync-readme-to-ecr-public.sh
```diff
@@ -0,0 +1,91 @@
...
+    fi
+}
+
+assert_not_empty region
+assert_not_empty repo_root
+assert_not_empty version
+
+#################################################
+
+if ! (git --no-pager diff --name-only HEAD^ HEAD | grep 'README.md' >/dev/null); then
```
snay2, line 61: The latent assumption here seems to be that the README, if it was modified at all since the prior release, will have been modified _at least_ in the most recent commit. In our current release process, that is generally true, but it is not obvious unless we read the script itself (perhaps a bit of a leaky abstraction).

This is probably okay, since a failure in which the README was modified (but not in the latest commit) will likely be fixed during the next release, even if we don't notice the failure at the time. Just wanted to clarify and see if you agree about that it is safe to make this assumption, rather than engineering a more robust solution.

Makefile
cjerad: It's just meant as a convenience.  The dev setup guide was updated to use the new target but I missed adding it to this commit.  Nice catch 👍

scripts/sync-readme-to-ecr-public.sh
cjerad: I copied this logic from the v1 script, but I agree with your point.  I'll update it to use the latest and previous release tags.

snay2 (approved):
LGTM

Issue #635, opened 2022-05-04
Emit pod events on node drain 


dennisme:
**Describe the feature**
A concise description of the feature and desired behavior.

On node drain the pods running on said node should emit events. These events can then be used to monitor at the pod level.  Without this you have to know which nodes a specific pod was running on then snag the node events. This feature would prevent that recursive lookup.

jillmon:
@dennisme, are you still working on this issue? The PR you posted was recently closed for inactivity. If you are still interested in making the change please reopen.

snay2:
Released in NTH `v1.18.0` (Helm chart version `v0.20.0`).

PR #631, opened 2022-04-21
Dennisme/rebase pod events

dennisme:
**Issue #635, if available:**

**Description of changes:**
This PR rebases an [earlier PR](https://github.com/newrelic-forks/aws-node-termination-handler/pull/1) on our fork by @dthorsen that adds  pod eventing to draining. We have been running this for a few months and it does help a ton with visibility. 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #411, opened 2021-04-15
feat: ability to emit Kubernetes events

trutx:
Fixes https://github.com/aws/aws-node-termination-handler/issues/414

Description of changes:

This PR adds the ability to emit Kubernetes events when interruption signals are received and when actions like cordon or drain are taken on nodes.

Kubernetes events are very useful for monitoring. At New Relic we push all cluster events using [an eventrouter](https://github.com/newrelic/nri-kube-events) up to [New Relic One](https://one.newrelic.com/) and then we create alerts and dashboards based on such events. This PR makes NTH emit a Kubernetes event every time a signal is received from AWS and also every time an action is taken on the node NTH is running on, regardless if it ends in success or failure.

The `involvedObject` for these events is the Kubernetes node NTH runs on, so they are published in the `default` namespace. To get them run `kubectl get events --field-selector source=aws-node-termination-handler`.

Additionally annotations can be attached to each event. Such annotations are useful for filtering and discovery. I.e. each event can be stamped with an annotation stating the cluster name the event belongs to, the team owning the cluster, etc. These annotations can later be used in the New Relic UI to filter out events belonging to a specific subset via a [NRQL query](https://docs.newrelic.com/docs/query-your-data/nrql-new-relic-query-language/get-started/introduction-nrql-new-relics-query-language/), or can be used to filter the results in a `kubectl get events` query piped to a processor like `jq` or `yq`.

PR #703, opened 2022-10-13
feat: emit pod events on drain

trutx:
**Issue #635:**

**Description of changes:**

This is a followup of @dennisme 's PR https://github.com/aws/aws-node-termination-handler/pull/631 which in turn was a followup of @dthorsen 's work. All credits should go to them. The PR was closed for being stale with some pending questions and requests. In this PR I'll just try to satisfy these and try to get it finally merged.

I'll comment the code exactly where comments or requests were in the original PR, creating links to each original question and mentioning the author, and then I'll try to answer the question or satisfy the request.

I'm also adding some minor changes to the PR: deprecating `io/ioutil` in favor of `os`.

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -64,6 +65,13 @@ const (
 	maxTaintValueLength = 63
 )
 
+const (
+	// PodEvictReason is the event reason emitted for Pod evictions during node drain
+	PodEvictReason = "PodEviction"
+	// PodEvictMsg is the event message emitted for Pod evictions during node drain
+	PodEvictMsg = "Pod evicted due to node drain"
+)
+
 var (
 	maxRetryDeadline      time.Duration = 5 * time.Second
 	conflictRetryInterval time.Duration = 750 * time.Millisecond
@@ -95,7 +103,7 @@ func NewWithValues(nthConfig config.Config, drainHelper *drain.Helper, uptime up
 }
 
 // CordonAndDrain will cordon the node and evict pods based on the config
-func (n Node) CordonAndDrain(nodeName string, reason string) error {
+func (n Node) CordonAndDrain(nodeName string, reason string, recorder recorderInterface) error {
 	if n.nthConfig.DryRun {
 		log.Info().Str("node_name", nodeName).Str("reason", reason).Msg("Node would have been cordoned and drained, but dry-run flag was set.")
 		return nil
@@ -113,6 +121,25 @@ func (n Node) CordonAndDrain(nodeName string, reason string) error {
 	node, err := n.fetchKubernetesNode(nodeName)
 	if err != nil {
 		return err
+	}
+	// Emit events for all pods that will be evicted
+	if recorder != nil {
+		pods, err := n.fetchAllPods(nodeName)
+		if err == nil {
+			for _, pod := range pods.Items {
+				podRef := &corev1.ObjectReference{
+					Kind:      "Pod",
+					Name:      pod.Name,
+					Namespace: pod.Namespace,
+				}
+				annotations := make(map[string]string)
+				annotations["node"] = nodeName
+				for k, v := range pod.GetLabels() {
+					annotations[k] = v
+				}
+				recorder.AnnotatedEventf(podRef, annotations, "Normal", PodEvictReason, PodEvictMsg)
+			}
+		}
 	}
 	err = drain.RunNodeDrain(n.drainHelper, node.Name)
 	if err != nil {
@@ -799,4 +826,8 @@ func filterPodForDeletion(podName string) func(pod corev1.Pod) drain.PodDeleteSt
 		}
 		return drain.MakePodDeleteStatusOkay()
 	}
+}
+
+type recorderInterface interface {
+	AnnotatedEventf(object runtime.Object, annotations map[string]string, eventtype, reason, messageFmt string, args ...interface{})
 }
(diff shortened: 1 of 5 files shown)


pkg/node/node.go
```diff
@@ -64,6 +65,13 @@ const (
 	maxTaintValueLength = 63
 )
 
+const (
+	// PodEvictReason is the event reason emitted for Pod evictions during node drain
+	PodEvictReason = "PodEviction"
+	// PodEvictMsg is the event message emitted for Pod evictions during node drain
+	PodEvictMsg = "Pod evicted due to node drain"
```
trutx, line 72: @snay2 asked https://github.com/aws/aws-node-termination-handler/pull/631/files#r868346527

> Is this sufficient detail for the event? Do you need it to specify that node-termination-handler initiated the drain in response to a particular AWS event (Spot ITN, ASG event, etc.)?

I'm modifying the `PodEvictMsg` to `PodEvictMsgFmt` that will take a string parameter. Such string will be the `nodeName`, so the user will be able to relate a pod eviction to the node where the pod was running in. Then it will be possible to look for that node's events and see the AWS event that caused it.

pkg/node/node.go
```diff
@@ -800,3 +827,7 @@ func filterPodForDeletion(podName string) func(pod corev1.Pod) drain.PodDeleteSt
 		return drain.MakePodDeleteStatusOkay()
 	}
 }
+
+type recorderInterface interface {
+	AnnotatedEventf(object runtime.Object, annotations map[string]string, eventtype, reason, messageFmt string, args ...interface{})
```
trutx, line 832: @AustinSiu asked https://github.com/aws/aws-node-termination-handler/pull/631/files#r868551337

> nit: `eventtype` -> `eventType`

Fixed.

pkg/node/node_test.go
```diff
@@ -61,7 +63,10 @@ func TestDryRun(t *testing.T) {
 	tNode, err := node.New(config.Config{DryRun: true})
 	h.Ok(t, err)
 
-	err = tNode.CordonAndDrain(nodeName, "cordonReason")
+	fakeRecorder := record.NewFakeRecorder(100)
```
trutx, line 66: @brycahta asked https://github.com/aws/aws-node-termination-handler/pull/631/files#r868317617

> * Make this const please
> * Is it necessary for the buffer size to be this large?

A `record.FakeRecorder` object can't be constant. We could use a global var but we could incur in concurrency problems when testing, I think it's better to just instantiate a new var scoped to every unit test.

Regarding the buffer size, I lowered it to 10.

pkg/node/node_test.go
```diff
@@ -98,6 +103,7 @@ func TestNewFailure(t *testing.T) {
 }
 
 func TestDrainSuccess(t *testing.T) {
+	controllerBool := true
```
trutx, line 106: @brycahta asked https://github.com/aws/aws-node-termination-handler/pull/631/files#r868318462
> update to be more descriptive, ex: `isOwnerController`

Fixed

pkg/node/node_test.go
```diff
@@ -106,14 +112,49 @@ func TestDrainSuccess(t *testing.T) {
...
+	h.Ok(t, err)
+
+	fakeRecorder := record.NewFakeRecorder(100)
+
 	tNode := getNode(t, getDrainHelper(client))
-	err = tNode.CordonAndDrain(nodeName, "cordonReason")
+	err = tNode.CordonAndDrain(nodeName, "cordonReason", fakeRecorder)
+
 	h.Ok(t, err)
+	close(fakeRecorder.Events)
```
trutx, line 142: @brycahta asked https://github.com/aws/aws-node-termination-handler/pull/631/files#r868327548

> Is there a reason to explicitly `close` this channel? If so, please update and use `defer` in case `h.Ok` fails.

If the channel is not closed, the upcoming `for event := range fakeRecorder.Events` loops forever. And if we use `defer` here, the test panics because `defer` attempts to close an already closed channel.

I have added two `defer` in the other two unit tests where the channel is created and never closed nor read. In this case here I agree that a failure in `h.Ok` would leave the channel open so I'm closing it before the assertion.

pkg/observability/k8s-events.go
```diff
@@ -113,7 +113,7 @@ func InitK8sEventRecorder(enabled bool, nodeName string, sqsMode bool, nodeMetad
 	}
 
 	broadcaster := record.NewBroadcaster()
-	broadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: clientSet.CoreV1().Events("default")})
+	broadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: clientSet.CoreV1().Events("")})
```
trutx, line 116: @snay2 asked https://github.com/aws/aws-node-termination-handler/pull/631/files#r868349413

> Can you provide some context on this change? What is the difference in behavior between `"default"` and `""`?

Haven't checked the code but I'm fairly sure they're equivalent. In fact I that `"default"` comes from my previous PR https://github.com/aws/aws-node-termination-handler/pull/411/files#diff-204929b469b9bbc88ae1116388da4e13f00f385c77c1db140d842371ab7e79f6R116 and then @dthorsen changed it to `""`. Maybe he can answer your question.

trutx:
Tests passed locally. Is there any way to relaunch them?

pkg/node/node.go
```diff
@@ -114,6 +122,25 @@ func (n Node) CordonAndDrain(nodeName string, reason string) error {
...
+					Kind:      "Pod",
+					Name:      pod.Name,
+					Namespace: pod.Namespace,
+				}
+				annotations := make(map[string]string)
+				annotations["node"] = nodeName
+				for k, v := range pod.GetLabels() {
+					annotations[k] = v
+				}
+				recorder.AnnotatedEventf(podRef, annotations, "Normal", PodEvictReason, PodEvictMsg)
```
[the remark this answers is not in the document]
snay2: Looks good. For reference, the two options are "normal" and "warning", so "normal" is the right choice here. Source: https://github.com/kubernetes/api/blob/v0.25.3/core/v1/types.go#L5755-L5761

snay2 (changes requested):
This looks good to me, other than I'd like to see the buffer size use a constant in the unit tests to ease readability.

pkg/node/node_test.go
trutx: Ah gotcha. Makes sense, just pushed a commit to create and use the `recorderBufferSize` const.

snay2 (approved):

Issue #575, opened 2022-02-07
NTH doesn't complete lifecycle hooks for instances that failed to join cluster


gabegorelick:
**Describe the bug**

I have ASG lifecycle hooks backed by NTH running in queue processor mode. If a node fails to join the cluster, our deploy infrastructure will eventually terminate it. However, NTH refuses to complete lifecycle hooks for instances that don't belong to the cluster. That means that the lifecycle hooks time out. 

This is despite the fact that NTH pods are running on other, healthy nodes in the cluster (I would totally understand if NTH failed to complete lifecycle hooks because NTH wasn't running 😄).

**Steps to reproduce**

1. Configure NTH in queue processor mode
2. Set up ASG lifecycle hooks
3. Deploy everything with CloudFormation `AutoScalingReplacingUpdate` so that new nodes are spun up in a new ASG that gets deleted upon deploy failure
4. Deploy an update that causes nodes to fail to join the cluster, e.g. remove the cluster info from your LaunchTemplate
5. Have CloudFormation roll back, or manually delete the ASG
6. ASG deletion triggers lifecycle hooks
7. NTH complains that "node 'XXXX' not found in the cluster"
8. ASG lifecycle hooks time out eventually and instances eventually terminate 

**Expected outcome**

I understand that NTH shouldn't try to drain or cordon nodes that don't belong to a cluster, but it should still complete any lifecycle hooks.

**Application Logs**
```
{"level":"info","event":{"EventID":"ec2-state-change-event-XXX","Kind":"SQS_TERMINATE","Description":"EC2 State Change event received. Instance i-XXX went into shutting-down at 2022-02-07 16:23:57 +0000 UTC \n","State":"","AutoScalingGroupName":"XXX","NodeName":"XXX","NodeLabels":null,"Pods":null,"InstanceID":"i-XXX","IsManaged":true,"StartTime":"2022-02-07T16:23:57Z","EndTime":"0001-01-01T00:00:00Z","NodeProcessed":false,"InProgress":false},"time":"2022-02-07T16:23:58Z","message":"Adding new event to the event store"}
{"level":"info","event-id":"ec2-state-change-event-XXX","kind":"SQS_TERMINATE","node-name":"ip-10-0-143-61.ec2.internal","instance-id":"i-XXX","time":"2022-02-07T16:23:58Z","message":"Requesting instance drain"}
[...]

PR #705, opened 2022-10-17
proceed lifecycle when node is not a part of the k8s cluster

gietschess:
**Issue #575**

**Description of changes:**
Possible way to solve the mentioned issue terminating instances which are not part of the k8s cluster.
Added flag `handleFailedInstances` to change the behaviour when the node which got the termination signal is not a cluster member. If set to `true` we proceed as the node is cordoned (and drained, depending on the arguments).

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -355,9 +355,9 @@ func drainOrCordonIfNecessary(interruptionEventStore *interruptioneventstore.Sto
 	}
 
 	if nthConfig.CordonOnly || (!nthConfig.EnableSQSTerminationDraining && drainEvent.IsRebalanceRecommendation() && !nthConfig.EnableRebalanceDraining) {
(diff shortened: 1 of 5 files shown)


snay2:
Thank you for the PR. Unfortunately, we are not inclined to accept it at this time, for two reasons. 

1) I [elaborated on the original issue](https://github.com/aws/aws-node-termination-handler/issues/575#issuecomment-1285776965) that we do not think NTH should complete lifecycle hooks in cases like this. Since NTH cannot know why the node failed to join the cluster, we don't want to incorrectly terminate valid customer workloads by completing lifecycle hooks when NTH does not have complete information about the instance.
2) Your solution is elegant in that it allows the customer to specify whether they want this behavior, so if they are okay with NTH masking failures from such misconfigurations, they can enable the option. However, we are trying to limit the proliferation of new options in NTH v1 because we are actively working on version 2, slated for beta release later this year.

PR #707, opened 2022-10-18
🥑🤖 v2.0.0-alpha release prep

cjerad:
🥑🤖 Auto-generated PR for v2.0.0-alpha release. Updating release versions in repo.

### DIFF
--- a/charts/aws-node-termination-handler-2/Chart.yaml
+++ b/charts/aws-node-termination-handler-2/Chart.yaml
@@ -3,7 +3,7 @@ name: aws-node-termination-handler-2
 description: A Helm chart for aws-node-termination-handler, an open-source component for gracefully handling termination events for node hosted in AWS.
 type: application
 version: "0.1.0"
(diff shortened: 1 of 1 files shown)


snay2 (approved):

stevehipwell:
@cjerad I've just had a very quick look at the Helm chart for v2 and it jumped out at me that you've added the CRD into the `templates` folder rather than the `crds` folder. If you're interested I'd be happy to take a look at the chart in more detail (I refactored the v1 chart)?

stevehipwell:
@cjerad the docs are specifically saying that you should use the `crds` folder and that Helm will no longer template CRD; this was an anti-pattern for Helm v2 and was explicitly called out as not to be used by Helm v3 and the v2 chart specification.

It not worth having Helm install the CRDs unless you're just giving a chart a quick test, the `crds` folder allows for this use case and also acts as the definitive source for the correct CRD to install via an alternate method.

PR #714, opened 2022-10-20
set git identity in sync-to-aws-eks-charts.sh

cjerad:
**Issue #, if available:**

Git user identity was not set in the sync-to-aws-eks-charts.sh script.

**Description of changes:**

Set the user identity for the local repo.

### DIFF
--- a/scripts/sync-to-aws-eks-charts.sh
+++ b/scripts/sync-to-aws-eks-charts.sh
@@ -153,13 +153,16 @@ cp -R "${helm_chart_path}/" "${eks_charts_nth_path}/"
 
 echo -e "🥑 Commit updates"
 
+git config --local user.name "ec2-bot 🤖"
+git config --local user.email "ec2-bot@users.noreply.github.com"
+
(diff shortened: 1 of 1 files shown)


scripts/sync-to-aws-eks-charts.sh
```diff
@@ -153,13 +153,16 @@ cp -R "${helm_chart_path}/" "${eks_charts_nth_path}/"
 
 echo -e "🥑 Commit updates"
 
+git config --local user.name "ec2-bot 🤖"
+git config --local user.email "ec2-bot@users.noreply.github.com"
```
snay2, line 157: This is to cause git to set "committer" and not just "author" on the commit, yes?

snay2 (approved):

scripts/sync-to-aws-eks-charts.sh
cjerad: The `git commit` command errored because the identity was not set.  Now that it *is* set, there is no need to explicitly specify the author of the commit.

PR #715, opened 2022-10-21
chore: add annotations to events in SQS mode

trutx:
**Description of changes:**

This PR enriches Kubernetes events when running in SQS mode by adding annotations to such events, so they can later be used to get information about the node involved in the event, create filters, etc. Similar to what was done in #411 

I have decided to add almost the same annotations the events in IMDS mode have, but I'd like some thoughts here because most of the information is gathered from node labels so we could simply attach all node labels as tags instead, or even do the two approaches simultaneously.

The only annotation I'm not adding here is `instance-life-cycle` because that's something that can be easily obtained from the IMDS endpoint but it's not present in the Kubernetes `Node` object, so it would have to be fetched from the AWS API instead (doable of course, but at the cost of another AWS API call). Again, I don't have a strong opinion on either option so thoughts welcome.

I intend to add a unit test (similar to `test/e2e/spot-interruption-test-events-on` once the set of annotations the events will have is decided. 

Also, I will submit another PR to update the documentation once this PR and also #703 and #706 land. At New Relic we are running an image containing the changes in all 3 PRs combined and it works great.

### DIFF
--- a/pkg/observability/k8s-events.go
+++ b/pkg/observability/k8s-events.go
@@ -14,6 +14,7 @@
 package observability
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
@@ -22,7 +23,10 @@ import (
 	"github.com/aws/aws-node-termination-handler/pkg/monitor/scheduledevent"
 	"github.com/aws/aws-node-termination-handler/pkg/monitor/spotitn"
 	"github.com/aws/aws-node-termination-handler/pkg/monitor/sqsevent"
+	"github.com/rs/zerolog/log"
 	corev1 "k8s.io/api/core/v1"
+	kErr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/observability/k8s-events.go
```diff
@@ -81,8 +87,8 @@ func InitK8sEventRecorder(enabled bool, nodeName string, sqsMode bool, nodeMetad
 	}
 
 	annotations := make(map[string]string)
+	annotations["account-id"] = nodeMetadata.AccountId
 	if !sqsMode {
-		annotations["account-id"] = nodeMetadata.AccountId
 		annotations["availability-zone"] = nodeMetadata.AvailabilityZone
 		annotations["instance-id"] = nodeMetadata.InstanceID
 		annotations["instance-life-cycle"] = nodeMetadata.InstanceLifeCycle
```
snay2, line 94: I agree with your decision to leave `instance-life-cycle` out of the annotations for SQS mode, since including it would require an additional AWS API call. If it turns out customers do want it later, we can discuss adding it in.

snay2 (approved):

pkg/observability/k8s-events.go
trutx: We can also add a new node-labels-to-event-annotations feature. Similar to the extra annotations we already have, but in this case the user would define a set (or even all?) of `Node` object labels that NTH would add to the event annotations map. Then the user could label all spot/ondemand instances they way they see fit (or any other arbitrary label FWIW) and instruct NTH to add such label as event annotation. NTH calls the k8s API anyway to get the `Node` object, so this extra feature would come for free in terms of remote API calls.

snay2:
Released in NTH `v1.18.0` (Helm chart version `v0.20.0`).

PR #716, opened 2022-11-01
update release workflow

cjerad:
**Issue #, if available:**

By default actions/checkout@v2 does not fetch git history.  Release script then fails when it attempts to look through the git history.

**Description of changes:**

* Configure actions/checkout@v2 to fetch git history
* Silence unnecessary script output


snay2 (approved):
lgtm

Issue #638, opened 2022-05-13
Migrate from `PodSecurityPolicy` to Pod Security Admission


snay2:
**Describe the bug**
When running the end-to-end tests on Kubernetes server version 1.23, I get the following warning:

```
policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+
```

**Steps to reproduce**
Run the following command on your local machine from the root of the repo:

```
test/k8s-local-cluster-test/run-test -v 1.23 -d
```

Then wait for it to build the image and start the cluster. After the first test starts running (e.g., "Running assertion script asg-lifecycle-sqs-test"), you'll see the warning above.

**Expected outcome**
We need to add support for Pod Security Admission before we can support Kubernetes server 1.25 ([expected release date August 2022](https://github.com/kubernetes/sig-release/tree/master/releases/release-1.25#summary)). However, if we wish to maintain support for the most recent 6 Kubernetes versions (which would include 1.20 and 1.21 at that time), we will need to keep the existing PodSecurityPolicy specs or use a third-party solution, because support for built-in Pod Security Admission began in 1.22.

The migration guide is here: https://kubernetes.io/docs/tasks/configure-pod-container/migrate-from-psp/

`PodSecurityPolicy` is used in several places throughout the repository: https://github.com/aws/aws-node-termination-handler/search?q=PodSecurityPolicy

**Application Logs**
See above.

**Environment**

* NTH App Version: 1.16.3
* NTH Mode (IMDS/Queue processor): N/A
* OS/Arch: MacOS 12.3.1
* Kubernetes version: 1.23
* Installation method: Source code

snay2:
Kubernetes v1.25 was released 3 weeks ago on 2022-08-23: https://kubernetes.io/releases/#release-v1-25

cjerad:
Included in the v1.19.0 release of AWS Node Termination Handler.

sernst:
Was the intention only to fix this in the Helm chart? Because the ` all-resources.yaml` and ` all-resources-queue-processor.yaml` for the v1.19.0 release here:

https://github.com/aws/aws-node-termination-handler/releases/tag/v1.19.0

still have those values. I see the conditional here in the Helm templates:

https://github.com/aws/aws-node-termination-handler/blob/main/config/helm/aws-node-termination-handler/templates/psp.yaml#L1

Given the number of historical versions that already support the pod security admissions as mentioned by @snay2 

[...]

PR #294, opened 2020-11-17
add prometheus helm chart for e2e tests

brycahta:
Issue #, if available:
* e2e tests will fail for updated [prometheus helm repo](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#prerequisites) for k8s 1.14 and 1.15

Description of changes:
* Moved chart from [old helm repo](https://github.com/helm/charts/tree/master/stable/prometheus-operator) and took out what wasn't necessary to get tests passing

Testing:
* `make e2e-test` passes

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #723, opened 2022-11-11
Remove bespoke Prometheus helm chart and use the latest public release instead

snay2:

**Description of changes:**
This more or less reverts https://github.com/aws/aws-node-termination-handler/pull/294.

The prior version of the Prometheus was `19.1.0`. I am pinning the new version to `41.7.4`, the latest release as of today, because it works and should let us have stable tests into the future.

Besides simplifying our ongoing maintenance, this will also serve to make deprecating `PodSecurityPolicy` easier (#638), because this helm chart has quite a few PodSecurityPolicies in it.

**Testing:**

I verified the Prometheus test before and after this change by comparing the test output. The test still passes with the newer, unmodified release of Prometheus.

This is the command to run just the Prometheus end-to-end test:

```bash
./test/k8s-local-cluster-test/run-test -b e2e-test -d -a test/e2e/prometheus-metrics-test
```


cjerad (approved):
lgtm 👍

Issue #732, opened 2022-11-28
Crash with exit code 2 in Queue Processing mode and IMDSv2 enabled on EC2


peteroruba:
**Describe the bug**
Process terminates with exit code 2

**Steps to reproduce**
Run in Queue Processing mode and enable IMDSv2 on EC2 instance

**Expected outcome**
IMDS endpoint is being ignored and NTH processes events via SQS Queue.

**Application Logs**
2022/11/28 14:26:29 WRN Deprecated argument "managed-asg-tag" and the replacement argument "managed-tag" was provided. Using the newer argument "managed-tag"
2022/11/28 14:26:29 WRN Deprecated argument "check-asg-tag-before-draining" and the replacement argument "check-tag-before-draining" was provided. Using the newer argument "check-tag-before-draining"
2022/11/28 14:26:29 INF Starting to serve handler /healthz, port 8080
2022/11/28 14:26:33 INF Metadata response status code: 401. Body: 
2022/11/28 14:26:33 INF aws-node-termination-handler arguments: 
	dry-run: false,
	node-name: ip-xx-xx-xx-xx.eu-central-1.compute.internal,
	pod-name: aws-node-termination-handler-5c45ff9756-2jlcv,
	metadata-url: http://169.254.169.254,
	kubernetes-service-host: 172.20.0.1,
	kubernetes-service-port: 443,
	delete-local-data: true,
	ignore-daemon-sets: true,
	pod-termination-grace-period: -1,
	node-termination-grace-period: 120,
	enable-scheduled-event-draining: true,
	enable-spot-interruption-draining: true,
	enable-sqs-termination-draining: true,
	enable-rebalance-monitoring: true,
	enable-rebalance-draining: true,
	metadata-tries: 3,
	cordon-only: false,
	taint-node: false,
	taint-effect: NoSchedule,
	exclude-from-load-balancers: false,
	json-logging: false,
	log-level: info,
	webhook-proxy: ,
	webhook-headers: <not-displayed>,
	webhook-url: ,
	webhook-template: <not-displayed>,
	uptime-from-file: ,
	enable-prometheus-server: false,
	prometheus-server-port: 9092,
	emit-kubernetes-events: false,
	kubernetes-events-extra-annotations: ,
	aws-region: eu-central-1,
	queue-url: https://sqs.eu-central-1.amazonaws.com/xxxxxx/node-termination-notification,
	check-tag-before-draining: true,
	managed-tag: aws-node-termination-handler/managed,
[...]

PR #735, opened 2022-11-30
Completely ignore Instance Metadata when in SQS Queue mode.

peteroruba:
**Issue #, if available:**
Fixes https://github.com/aws/aws-node-termination-handler/issues/732

When IMDSv2 is enabled and aws-note-termination-handler is ran inside a container accessing IMDSv2 is not possible due to the IMDSv2 HTTP endpoint's default hop limit of 1. This makes the process exit.

**Description of changes:**
Completely disables the communication with the IMDS endpoint when in SQS queue mode.

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -113,11 +113,12 @@ func main() {
 		nthConfig.Print()
 		log.Fatal().Err(err).Msg("Unable to instantiate probes service,")
 	}
+        imdsDisabled := nthConfig.EnableSQSTerminationDraining
 
 	imds := ec2metadata.New(nthConfig.MetadataURL, nthConfig.MetadataTries)
 
 	interruptionEventStore := interruptioneventstore.New(nthConfig)
-	nodeMetadata := imds.GetNodeMetadata()
+	nodeMetadata := imds.GetNodeMetadata(!imdsDisabled)
 	// Populate the aws region if available from node metadata and not already explicitly configured
 	if nthConfig.AWSRegion == "" && nodeMetadata.Region != "" {
 		nthConfig.AWSRegion = nodeMetadata.Region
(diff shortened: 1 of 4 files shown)


pkg/ec2metadata/ec2metadata.go
```diff
@@ -325,30 +325,43 @@ func retry(attempts int, sleep time.Duration, httpReq func() (*http.Response, er
...
+	    metadata.InstanceID        = ""
+	    metadata.InstanceLifeCycle = ""
+	    metadata.InstanceType      = ""
+	    metadata.PublicHostname    = ""
+	    metadata.PublicIP          = ""
+	    metadata.LocalHostname     = ""
+	    metadata.LocalIP           = ""
+	    metadata.AvailabilityZone  = ""
+	    metadata.Region            = ""
+        }
```
snay2, line 364: Is this `else` branch necessary? When `imdsDisabled=false`, `metadata` should still be empty, since no properties were set. When you were testing, did you discover this was required?

cmd/node-termination-handler.go
```diff
@@ -113,11 +113,12 @@ func main() {
 		nthConfig.Print()
 		log.Fatal().Err(err).Msg("Unable to instantiate probes service,")
 	}
+        imdsDisabled := nthConfig.EnableSQSTerminationDraining
 
 	imds := ec2metadata.New(nthConfig.MetadataURL, nthConfig.MetadataTries)
 
 	interruptionEventStore := interruptioneventstore.New(nthConfig)
-	nodeMetadata := imds.GetNodeMetadata()
+	nodeMetadata := imds.GetNodeMetadata(!imdsDisabled)
```
snay2, line 121: Given that both the variable and the function parameter are called "disabled", shouldn't this line be the following?

```suggestion
	nodeMetadata := imds.GetNodeMetadata(imdsDisabled)
```

snay2 (changes requested):
This unit test needs an update: https://github.com/aws/aws-node-termination-handler/blob/main/pkg/ec2metadata/ec2metadata_test.go#L583

Also, if possible, please add a new unit test to verify that IMDS does not get called when calling `GetNodeMetadata()` in Queue Processor mode.

pkg/ec2metadata/ec2metadata.go
peteroruba: He complained about metadata being uninitialized so I've set it to at least empty string values.

peteroruba:
Unfortunately I am not able to run the tests locally, as localstack doesn't come up properly:

🥑 Using localstack pod localstack-7f5f94f966-qbdr5

Could not connect to the endpoint URL: "http://localhost:4597/"
command terminated with exit code 255
error: You must provide one or more resources by argument or filename.
Example resource specifications include:
   '-f rsrc.yaml'
   '--filename=rsrc.json'
   '<resource> <name>'
   '<resource>'
⏰ Took 385sec
❌ NTH Integration Test FAILED e2e-test-bed7523d! ❌
🥑 Deleting k8s cluster using "kind"
Deleting cluster "e2e-test-bed7523d" ...
make: *** [e2e-test] Error 1

pkg/ec2metadata/ec2metadata.go
snay2: Who is "he"?

peteroruba: The aws-node-termination-handler process. I ran it both on AWS EKS and on my workstation's local k8s cluster.

cmd/node-termination-handler.go
peteroruba: To clarify: GetNodeMetadata() does get called in queue mode with the current code as per this MR. Unfortunately my Gelang skills are not sufficient to come up with an elegant refactoring that truly separates the code parts.

snay2 (changes requested):
Three changes still needed:

- [ ] Update the `NodeMetadata` initialization logic (see [this comment](https://github.com/aws/aws-node-termination-handler/pull/735/files#r1039902511))
- [ ] Add a unit test for `GetNodeMetadata()` when IMDS is disabled
- [ ] Update the README to indicate that we no longer query IMDS on startup when using Queue Processor mode

I made these changes in a commit on my fork: https://github.com/snay2/aws-node-termination-handler/commit/d2b225661e07a1baecdaa61efe48170665ff4836

You can cherry-pick that commit onto your branch and update the PR. Otherwise, I can push the commit directly to your branch, whichever you prefer.

peteroruba:
@snay2 I cherry-picked your commit. Your assistance on this matter is much appreciated.

snay2 (approved):
LGTM. Thanks for the PR!

PR #741, opened 2022-12-06
Enable scheduled event draining by default

johngmyers:
**Issue #, if available:**

N/A

**Description of changes:**

Changes the default for scheduled event draining to enabled.


snay2:
We're not inclined to take this change without a compelling reason for why we should change the default for all customers. Can you elaborate on your motivation?

johngmyers:
The motivation is that without the feature being enabled, the application code may not stop gracefully, take longer to recover full availability, or accidentally schedule work to nodes that are terminated due to a scheduled maintenance event.

Users should not have to make extra effort for systems to have reasonable behavior. The reasonable expectation of a user would be for pods on instances scheduled to be terminated for maintenance would be to have those pods be gracefully evicted, not summarily deleted with the instance.

snay2:
Thanks, that's a good argument.

In the past, scheduled event draining was not enabled by default because it was a new feature we added after spot ITN draining. But for a new customer installing NTH now, that historical reason for the defaults doesn't matter.

Since scheduled events come directly from EC2 like spot ITNs and do not require any customer configuration and are a definitive signal that the instance will eventually be terminated (unlike, e.g., rebalance recommendations which merely indicate that the instance is at higher risk of termination).

snay2 (approved):

cjerad:
Included in release v1.18.2

PR #743, opened 2022-12-10
Don't try to uncordon on startup in Queue Processor mode

johngmyers:
**Issue #, if available:**

N/A

**Description of changes:**

The documentation states:

> The Queue Processor Mode does not allow for fine-grained configuration of which events are handled through helm configuration keys.

The `.Values.enableScheduledEventDraining` helm configuration key does control one aspect of NTH in Queue Processor mode: whether upon startup it attempts to uncordon the node it is running on.

This appears to be a bug. There is no logical reason for NTH to uncordon the node it is running on when not in IMDS mode.

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -113,7 +113,7 @@ func main() {
 		nthConfig.Print()
 		log.Fatal().Err(err).Msg("Unable to instantiate probes service,")
 	}
-        imdsDisabled := nthConfig.EnableSQSTerminationDraining
+	imdsDisabled := nthConfig.EnableSQSTerminationDraining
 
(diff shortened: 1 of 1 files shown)


johngmyers:
@peteroruba you missed this one.

snay2 (approved):
Thanks for the fix!

This will also be important when we remove PodSecurityPolicy (#638), because we want Queue Processor mode to be able to run in a Baseline or Restricted namespace. One thing that would require NTH to run in a Privileged namespace is filesystem access, and this feature tries to read `/proc/uptime`. That will still be necessary for IMDS mode, but QP mode doesn't need it. Thanks for making that actually true. :)

PR #744, opened 2022-12-10
Announce deprecation of IMDS mode

johngmyers:
**Issue #, if available:**

None

**Description of changes:**

The `v2` branch of NTH removes IMDS mode. This change announces the deprecation of the mode.


snay2:
While it is true that NTHv2 does not support IMDS mode because of its architecture, we are not deprecating NTHv1 yet--it will continue to support IMDS mode.

PR #745, opened 2022-12-10
Remove unnecessary configuration in Queue Processor mode

johngmyers:
**Issue #, if available:**

None

**Description of changes:**

Remove configuration settings that don't affect Queue Processor mode (assuming #743 merges) from the Deployment template, so that they don't act as red herrings.

Document the settings that only apply to IMDS mode.


config/helm/aws-node-termination-handler/README.md
```diff
@@ -150,12 +150,12 @@ The configuration in this table applies to AWS Node Termination Handler in IMDS
...
 | `windowsTolerations`             | Override `daemonsetTolerations` for the Linux DaemonSet.                                                                                                                                                                                                      | `[]`                   |
-| `enableProbesServer`             | If `true`, start an http server exposing `/healthz` endpoint for probes.                                                                                                                                                                                      | `false`                |
+| `enableProbesServer`             | If `true`, start an http server exposing `/healthz` endpoint for probes. Only used in IMDS mode.                                                                                                                                                              | `false`                |
```
snay2, line 153: This is incorrect. The probes server can be exposed whether NTH runs as a daemonset or as a deployment and is not related to IMDS mode. Our default configurations here only enable it for the daemonset, not the deployment, but that is not a technical limitation.

Indeed, our release artifacts explicitly include an aggregated YAML configuration with probes server enabled in Queue Processor mode as a deployment (see the artifact `all-resources-queue-processor.yaml` from any v1 release). Here's the code that generates that file:

https://github.com/aws/aws-node-termination-handler/blob/c3e411b981db6911c668f979a07f25d431cbe073/scripts/generate-k8s-yaml#L61-L66

config/helm/aws-node-termination-handler/values.yaml
```diff
@@ -259,22 +259,22 @@ daemonsetTolerations:
 linuxTolerations: []
 windowsTolerations: []
 
-# If the probes server is running for the Daemonset
+# If the probes server is running for the Daemonset. Only used in IMDS mode.
```
snay2, line 262: To be more correct, let's remove "for the Daemonset", since the probes server can be used in a deployment and is not related to IMDS mode.

```suggestion
# If the probes server is running
```

snay2 (approved):

PR #746, opened 2022-12-10
Don't skip evicting pods with same name but in different namespace

johngmyers:
**Issue #, if available:**

None

**Description of changes:**

When draining a node, NTH skips evicting itself. It incorrectly skipped evicting pods in other namespaces that happened to have the same name.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -30,6 +30,7 @@ const (
 	dryRunConfigKey                         = "DRY_RUN"
 	nodeNameConfigKey                       = "NODE_NAME"
 	podNameConfigKey                        = "POD_NAME"
+	podNamespaceConfigKey                   = "NAMESPACE"
 	kubernetesServiceHostConfigKey          = "KUBERNETES_SERVICE_HOST"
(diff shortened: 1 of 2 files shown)


johngmyers:
The check failure appears to be a flake.

snay2:
Can you open an issue with more details about what is happening and why? I'm not sure I understand the usecase or how common it is. We try to be judicious about adding new configuration options to NTHv1 because it already has so many.

johngmyers:
The NTH Helm chart already passes in the NAMESPACE environment variable, but NTH doesn't reference it. Since the value is already available, correctness is preferable.

If it is desired to work less correctly without the setting, the condition could be written as `pod.Name == podName && (pod.Namespace == podNamespace || podNamespace == ""`

snay2:
That's good to know. If the environment variable is already in the Helm chart, then this seems like a reasonable thing to include.

I'm in favor of including the `|| podNamespace == ""` condition to improve backwards compatibility.

snay2 (approved):
LGTM

PR #748, opened 2022-12-13
remove event schema version checks

cjerad:
**Issue #, if available:**

Incoming event versions do not match expected schema versions.

**Description of changes:**

For now, remove the event version checks.  We may re-add the checks in the future.

### DIFF
--- a/pkg/event/asgterminate/v1/parser.go
+++ b/pkg/event/asgterminate/v1/parser.go
@@ -27,7 +27,6 @@ import (
 const (
 	source             = "aws.autoscaling"
 	detailType         = "EC2 Instance-terminate Lifecycle Action"
-	version            = "1"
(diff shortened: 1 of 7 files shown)


pkg/event/asgterminate/v1/parser.go
```diff
@@ -27,7 +27,6 @@ import (
 const (
 	source             = "aws.autoscaling"
 	detailType         = "EC2 Instance-terminate Lifecycle Action"
-	version            = "1"
```
jillmon, line 30: Are we just no longer depending on version number for anything?

cjerad: Temporarily. Need to look into why the events don't have the expected version number. Later, if it makes sense, we can restore the checks.

jillmon (approved):

Issue #685, opened 2022-09-08
All retries failed, unable to complete the uncordon after reboot workflow error


sushantsoni5392:
**Describe the bug**
Hi,

In the logs right after the NTH starts we can see errors frequently like below 
```
2022/09/08 08:18:46 ERR Error when trying to list Nodes w/ label, falling back to direct Get lookup of node error="Get \"https://172.20.0.1:443/api/v1/nodes?labelSelector=kubernetes.io%2Fhostname%3D%3Dip-10-45-5-107.eu-central-1.compute.internal\": dial tcp 172.20.0.1:443: i/o timeout"
2022/09/08 08:18:46 WRN All retries failed, unable to complete the uncordon after reboot workflow error="timed out waiting for the condition"
```

I wanted to understand if this error affects anything.

**Steps to reproduce**

**Expected outcome**
No errors

**Application Logs**
The log output when experiencing the issue.
```
2022/09/08 08:18:14 INF aws-node-termination-handler arguments:
	dry-run: false,
	node-name: ip-10-45-5-137.eu-central-1.compute.internal,
	pod-name: aws-node-termination-handler-866sr,
	metadata-url: http://169.254.169.254,
	kubernetes-service-host: 172.20.0.1,
	kubernetes-service-port: 443,
	delete-local-data: true,
	ignore-daemon-sets: true,
	pod-termination-grace-period: -1,
	node-termination-grace-period: 120,
	enable-scheduled-event-draining: true,
	enable-spot-interruption-draining: true,
	enable-sqs-termination-draining: false,
	enable-rebalance-monitoring: true,
	enable-rebalance-draining: false,
	metadata-tries: 3,
	cordon-only: false,
	taint-node: true,
	taint-effect: NoSchedule,
	exclude-from-load-balancers: false,
	json-logging: false,
	log-level: info,
	webhook-proxy: ,
	webhook-headers: <not-displayed>,
	webhook-url: ,
	webhook-template: <not-displayed>,
	uptime-from-file: /proc/uptime,
	enable-prometheus-server: false,
	prometheus-server-port: 9092,
	emit-kubernetes-events: false,
	kubernetes-events-extra-annotations: ,
	aws-region: eu-central-1,
	queue-url: ,
	check-asg-tag-before-draining: true,
	managed-asg-tag: aws-node-termination-handler/managed,
	assume-asg-tag-propagation: false,
	aws-endpoint: ,

[...]

PR #756, opened 2023-01-19
Error log fix

LikithaVemulapalli:
**Issue #, if available:** https://github.com/aws/aws-node-termination-handler/issues/685

**Description of changes:**

- As per the PR comments, modified log statement from Error to Warning as this doesn't affect any expected outcome.
- This error is mainly due to mismatch between Node labels as per the code and we already have a fallback mechanism.
-  Ran tests locally and all tests passed.

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -611,7 +611,7 @@ func (n Node) fetchKubernetesNode(nodeName string) (*corev1.Node, error) {
 	listOptions := metav1.ListOptions{LabelSelector: labels.Set(labelSelector.MatchLabels).String()}
 	matchingNodes, err := n.drainHelper.Client.CoreV1().Nodes().List(context.TODO(), listOptions)
 	if err != nil || len(matchingNodes.Items) == 0 {
-		log.Err(err).Msgf("Error when trying to list Nodes w/ label, falling back to direct Get lookup of node")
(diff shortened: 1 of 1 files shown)


pkg/node/node.go
```diff
@@ -611,7 +611,7 @@ func (n Node) fetchKubernetesNode(nodeName string) (*corev1.Node, error) {
 	listOptions := metav1.ListOptions{LabelSelector: labels.Set(labelSelector.MatchLabels).String()}
 	matchingNodes, err := n.drainHelper.Client.CoreV1().Nodes().List(context.TODO(), listOptions)
 	if err != nil || len(matchingNodes.Items) == 0 {
-		log.Err(err).Msgf("Error when trying to list Nodes w/ label, falling back to direct Get lookup of node")
+		log.Warn().Msgf("Unable to list Nodes w/ label, falling back to direct Get lookup of node")
```
cjerad, line 614: Should the output include the `err`?

```suggestion
		log.Warn().Err(err).Msgf("Unable to list Nodes w/ label, falling back to direct Get lookup of node")
```

LikithaVemulapalli: From the comments, I believe it would be more clear to keep it as a Warning and not include Error in the output!

cjerad (approved):
lgtm 👍

Issue #757, opened 2023-01-23
Spot interruption event logs incorrect time


aarongorka:
**Describe the bug**
When NTH receives a spot interruption notice and logs the interruption time, it logs the current time rather than the actual interruption time.

See also https://github.com/aws/aws-node-termination-handler/discussions/606

**Steps to reproduce**
  1. Set up NTH in Deployment/SQS mode
  2. Configure EventBridge to send spot termination notices
  3. Wait for a spot termination notice
  4. Observe log with "Spot Interruption event received"

**Expected outcome**
Spot interruption time should be roughly +2 minutes from when the event was received. Ideally we can use the actual interruption time as supplied by AWS rather than guestimating it though.

**Application Logs**
```
2023-01-23T00:18:31.086824792Z aws-node-termination-handler-6574867b74-z7l57 {"level":"info","event":{"EventID":"spot-itn-event-62363231313539382d666634352d343631612d646537642d346232353565353832386463","Kind":"SPOT_ITN","Monitor":"SQS_MONITOR","Description":"Spot Interruption event received. Instance i-099447d7bc6f9dde7 will be interrupted at 2023-01-23 00:18:29 +0000 UTC \n","State":"","AutoScalingGroupName":"eks-main-20220713005232834100000009","NodeName":"ip-10-9-139-214.ec2.internal","NodeLabels":null,"Pods":null,"InstanceID":"i-099447d7bc6f9dde7","ProviderID":"aws:///us-east-1b/i-099447d7bc6f9dde7","IsManaged":true,"StartTime":"2023-01-23T00:18:29Z","EndTime":"0001-01-01T00:00:00Z","NodeProcessed":false,"InProgress":false},"time":"2023-01-23T00:18:31Z","message":"Adding new event to the event store"}
2023-01-23T00:20:18.474299920Z aws-node-termination-handler-6574867b74-z7l57 {"level":"info","node_name":"ip-10-9-139-214.ec2.internal","reason":"Spot Interruption event received. Instance i-099447d7bc6f9dde7 will be interrupted at 2023-01-23 00:18:29 +0000 UTC \n","time":"2023-01-23T00:20:18Z","message":"Node successfully cordoned and drained"}
```

**Environment**

* NTH App Version: v1.18.2 (Helm chart version 0.18.5)
* NTH Mode (IMDS/Queue processor): Queue processor
* OS/Arch: Amazon Linux 2, x86
* Kubernetes version: 1.24
* Installation method: Helm

[...]

PR #762, opened 2023-01-27
Modified Spot ITN message

LikithaVemulapalli:
**Issue #, if available:** https://github.com/aws/aws-node-termination-handler/issues/757

**Description of changes:**

We are returning the EventBridge event start time, in our case when instance received spot interruption. Current time is returned when we cannot parse the time in the specific format and we get a warning message that `Unable to parse time as RFC3339 from event, using current time instead`, we cannot hardcode the time for spot interruption as it might take 2 or minutes to terminate. Modified the Spot Interruption event message to make it more clear.

### DIFF
--- a/pkg/monitor/sqsevent/spot-itn-event.go
+++ b/pkg/monitor/sqsevent/spot-itn-event.go
@@ -29,7 +29,7 @@ import (
 	"id": "1e5527d7-bb36-4607-3370-4164db56a40e",
 	"detail-type": "EC2 Spot Instance Interruption Warning",
 	"source": "aws.ec2",
-	"account": "123456789012",
+	"account": "<account_number>",
 	"time": "1970-01-01T00:00:00Z",
 	"region": "us-east-1",
 	"resources": [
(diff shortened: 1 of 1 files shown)


pkg/monitor/sqsevent/spot-itn-event.go
```diff
@@ -69,7 +69,7 @@ func (m SQSMonitor) spotITNTerminationToInterruptionEvent(event *EventBridgeEven
 		IsManaged:            nodeInfo.IsManaged,
 		InstanceID:           spotInterruptionDetail.InstanceID,
 		ProviderID:           nodeInfo.ProviderID,
-		Description:          fmt.Sprintf("Spot Interruption event received. Instance %s will be interrupted at %s \n", spotInterruptionDetail.InstanceID, event.getTime()),
+		Description:          fmt.Sprintf("Spot Interruption event received. Instance %s received interruption at %s \n", spotInterruptionDetail.InstanceID, event.getTime()),
```
cjerad, line 72: This might still be misinterpreted as "the instance was interrupted at".  How about the following?

```suggestion
		Description:          fmt.Sprintf("Spot Interruption notice for instance %s was sent at %s \n", spotInterruptionDetail.InstanceID, event.getTime()),
```

LikithaVemulapalli: Yes, this looks good to me! Will modify this one! Thanks for the suggestion :)

cjerad (approved):
lgtm 👍

PR #851, opened 2023-06-21
Upload k8s assets only on the second upload call

pdk27:
Upload k8s assets only on the second upload call


LikithaVemulapalli (approved):
/lgtm

PR #943, opened 2024-01-18
fix: exit when recv sig

rfyiamcool:
### summary

process exit when recv signal. exit 2 level loop, otherwise `wg.wait` will block ? 😁

```
	for range time.NewTicker(1 * time.Second).C {
		select {
		case <-signalChan:
			break  // 😅 the `break` only break select-loop, but don't break for-loop outside.
                 }
                 ...
}
```


rfyiamcool:
?

PR #999, opened 2024-05-09
fix: Reverted the removal of the lifecycle hook completion

### LINKED ISSUE #990: NTH queue handler behaviour broken by PR #940
**Describe the bug**
PR #940 removed the continue lifecycle action behaviour for ASGs which breaks the expected NTH behaviour; which is to signal to the ASG that the node can be deleted once it's drained.

If the PR author wanted to disable this functionality for their use case they should have provided configuration to do so, and only changed the default behaviour as part of a major version (if even wanted).

**Steps to reproduce**
Replace an ASG instance with instance refresh, the instance will not be terminated until the lifecycle hook times out.

**Expected outcome**
I'd expect the instance to be terminated by the ASG once NTH has completed draining it.

**Application Logs**
The log output when experiencing the issue.

**Environment**

* NTH App Version: v1.21.0
* NTH Mode (IMDS/Queue processor): Queue processor
* OS/Arch: n/a
* Kubernetes version: n/a
* Installation method: Helm

stevehipwell:
**Issue #, if available:**
- Fixes #990

**Description of changes:**
This PR reverts the removal of the lifecycle hook completions for ASG lifecycle events.

### DIFF
--- a/pkg/monitor/sqsevent/asg-lifecycle-event.go
+++ b/pkg/monitor/sqsevent/asg-lifecycle-event.go
@@ -95,6 +95,11 @@ func (m SQSMonitor) asgTerminationToInterruptionEvent(event *EventBridgeEvent, m
 	}
 
 	interruptionEvent.PostDrainTask = func(interruptionEvent monitor.InterruptionEvent, _ node.Node) error {
+		_, err = m.continueLifecycleAction(lifecycleDetail)
+		if err != nil {
+			return fmt.Errorf("continuing ASG termination lifecycle: %w", err)
+		}
+		log.Info().Str("lifecycleHookName", lifecycleDetail.LifecycleHookName).Str("instanceID", lifecycleDetail.EC2InstanceID).Msg("Completed ASG Lifecycle Hook")
 		return m.deleteMessage(message)
 	}
 


pkg/monitor/sqsevent/asg-lifecycle-event.go
```diff
@@ -95,6 +95,11 @@ func (m SQSMonitor) asgTerminationToInterruptionEvent(event *EventBridgeEvent, m
 	}
 
 	interruptionEvent.PostDrainTask = func(interruptionEvent monitor.InterruptionEvent, _ node.Node) error {
+		_, err = m.continueLifecycleAction(lifecycleDetail)
+		if err != nil {
+			return fmt.Errorf("continuing ASG launch lifecycle: %w", err)
```
LikithaVemulapalli, line 100: quick question: the error message seems confusing here, should it be terminate/ ASG lifecycle instead of launch in `asgTerminationToInterruptionEvent` method ? can you modify this to make it more clear as we have same error message in `createAsgInstanceLaunchEvent` method. I will approve this and will do NTH release either today or tomorrow once this PR is merged. Thanks :)

LikithaVemulapalli (commented):
Hello @stevehipwell, I appreciate the efforts for these changes. I left a comment, could you please take a look at it today so that I can merge this and do an NTH release soon. Thanks :)

pkg/monitor/sqsevent/asg-lifecycle-event.go
stevehipwell: Sorry, this should be fixed now.

LikithaVemulapalli (approved):
/lgtm, thanks for the quick fix. We are planning to do the release on Monday. Will update the original issue with release version 👍

Issue #1017, opened 2024-05-29
Expose option to not do an etcd quorum read when listing pods on a node


hieurender:
**Describe the feature**
The `fetchAllPods` function currently lists all pods in such a way that it requires etcd to do a quorum read out of all pods in the cluster, and pass that (potentially very large) list to the k8s apiserver, who then selects only the ones with the desired `spec.nodeName` field. In very large clusters, this can be slow operation that's arduous for etcd to execute.

A less expensive way of achieving the same thing would be to leverage the apiserver cache, which indexes pods by their node name. This can be done by passing a non-empty `ResourceVersion` in the `metav1.ListOptions`.

I would like the nthConfig to expose a flag to configure this behavior, with the status quo being the default, since most users seem to be happy with a consistent read.

The situation is very similar to [this issue](https://github.com/vectordotdev/vector/issues/16797) in the vector project, which was fixed by [this PR](https://github.com/vectordotdev/vector/pull/17095).

**Is the feature request related to a problem?**
A description of what the problem is. For example: I'm frustrated when [...]

**Describe alternatives you've considered**
A description of any alternative solutions or features you've considered.

LikithaVemulapalli:
Support for this issue is now merged, closing this issue. Thanks!

PR #1018, opened 2024-05-29
Allow using kube-apiserver cache for pod list requests

hieurender:
**Issue #, if available:** https://github.com/aws/aws-node-termination-handler/issues/1017

**Description of changes:** Exposes the configuration `use_apiserver_cache`, which sets `ResourceVersion: "0"` when listing pods on a node.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -88,6 +88,7 @@ const (
 	uptimeFromFileDefault                   = ""
 	workersConfigKey                        = "WORKERS"
 	workersDefault                          = 10
+	useAPIServerCache                       = "USE_APISERVER_CACHE"
 	// prometheus
 	enablePrometheusDefault   = false
 	enablePrometheusConfigKey = "ENABLE_PROMETHEUS_SERVER"
@@ -161,6 +162,7 @@ type Config struct {
 	UseProviderId                       bool
 	CompleteLifecycleActionDelaySeconds int
 	DeleteSqsMsgIfNodeNotFound          bool
+	UseAPIServerCacheToListPods         bool
 }
 
 // ParseCliArgs parses cli arguments and uses environment variables as fallback values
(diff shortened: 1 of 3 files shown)


LikithaVemulapalli (changes requested):
Hello @hieurender, thanks for opening a PR for the issue that you were facing. We added multiple configs earlier as per the requirements and were reluctant to adding additional configs to reduce complexity. We had seen some interest on this config from other users, we are happy to merge it. Could you please add some tests around the newly added config in `config_test.go`. Thanks

hieurender:
@LikithaVemulapalli added the test you asked for.

I also added a second commit, as even with the first commit we noticed NTH was still doing quorum-lists of pods. We traced the source to the call `drain.RunNodeDrain`, which calls `drainer.GetPodsForDeletion`, which does not set ResourceVersion=0 in its ListOptions: 

```
	podList := &corev1.PodList{}
	initialOpts := &metav1.ListOptions{
		LabelSelector: labelSelector.String(),
		FieldSelector: fields.SelectorFromSet(fields.Set{"spec.nodeName": nodeName}).String(),
		Limit:         d.ChunkSize,
	}
```

It was fairly straightforward to refactor so that we don't call this function.

LikithaVemulapalli:
@hieurender could you please check the test failures and fix them. We observed these test failures for your changes related to latest commit, we can only approve them if there are no issues for the existing functionality, once the tests passes will review the changes. Thanks.

hieurender:
I verified that `make e2e-test` passes locally 👍

LikithaVemulapalli (approved):
/lgtm, thanks for fixing :)

hieurender:
@LikithaVemulapalli are there guidelines on releasing new versions? I'd love to be able to move off my fork.

LikithaVemulapalli:
Hello @hieurender, we will be doing release in a timely basis every month, we will be doing release in mid July. Will update your original issue or this PR once this is included in the new release. Thanks

PR #1058, opened 2024-08-29
Misleading 404 logs in IMDS mode when querying EC2 Metadata without a public IP

xabinapal:
**Description of changes:**

While running NTH in IMDS mode on instances without an assigned public IP, I noticed that requests to certain EC2 Metadata endpoints, specifically `public-hostname` and `public-ipv4`, return a 404 status code.  This is the expected behavior, as documented [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#instancedata-data-categories). Although NTH does not fail and works as expected in this scenario, some log messages about these 404 status codes are displayed:

```
2024/08/29 13:42:34 INF Metadata response status code: 404. Body: <?xml version="1.0" encoding="iso-8859-1"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
		 "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
  <title>404 - Not Found</title>
 </head>
 <body>
  <h1>404 - Not Found</h1>
 </body>
</html>
```

These messages don't clarify the reason for the 404 codes, which can mislead users (like myself!) into thinking something is wrong. To address this, I've added an `allowMissing` parameter to the `GetMetadataInfo` function. When this flag is enabled, no message will be logged for 404 responses, and no error will be returned.

While this might not be the most elegant solution, I felt like any other alternative, such as implementing the options pattern, would be overkill and too complex for the limited scope of the function.

### DIFF
--- a/pkg/ec2metadata/ec2metadata.go
+++ b/pkg/ec2metadata/ec2metadata.go
@@ -194,7 +194,7 @@ func (e *Service) GetRebalanceRecommendationEvent() (rebalanceRec *RebalanceReco
 }
 
 // GetMetadataInfo generic function for retrieving ec2 metadata
-func (e *Service) GetMetadataInfo(path string) (info string, err error) {
+func (e *Service) GetMetadataInfo(path string, allowMissing bool) (info string, err error) {
 	metadataInfo := ""
 	resp, err := e.Request(path)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


xabinapal:
Could this be reviewed so it's not marked as stale by the bot?

LikithaVemulapalli:
Hello @xabinapal, will take a look at this PR. Thanks for your contribution :)

LikithaVemulapalli:
Hello @xabinapal, can you pull in the latest commit that will fix the lint test failure, then I will merge your change, thanks :)

LikithaVemulapalli (approved):
/lgtm

PR #1083, opened 2024-10-29
Prometheus add nodes gauge for SQS mode 

### LINKED ISSUE #785: Add metrics to show the number of nodes being tracked
**Describe the feature**
I'd like to know the number of nodes that a NTH instance is currently managing and have it exported as a metric.

**Is the feature request related to a problem?**
After the tagging changes we had a defect where the tags weren't being set on the instances causing our nodes to no longer be managed by NTH; this metric would have meant it was easy to see that it wasn't configured correctly.

**Describe alternatives you've considered**
n/a

phuhung273:
**Issue #, if available:**
Close #785 

**Description of changes:**
- Count nodes/instances being tracked. Eg:
`k get node` return 5
`aws ec2 describe-instances` return 2
**Identify 3 nodes no longer under NTH control**

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -120,10 +120,10 @@ func main() {
 		log.Fatal().Err(err).Msg("Unable to instantiate a node for various kubernetes node functions,")
 	}
 
-	metrics, err := observability.InitMetrics(nthConfig.EnablePrometheus, nthConfig.PrometheusPort)
-	if err != nil {
+	metrics, initMetricsErr := observability.InitMetrics(nthConfig.EnablePrometheus, nthConfig.PrometheusPort)
+	if initMetricsErr != nil {
 		nthConfig.Print()
-		log.Fatal().Err(err).Msg("Unable to instantiate observability metrics,")
+		log.Fatal().Err(initMetricsErr).Msg("Unable to instantiate observability metrics,")
 	}
 
 	err = observability.InitProbes(nthConfig.EnableProbes, nthConfig.ProbesPort, nthConfig.ProbesEndpoint)
@@ -217,7 +217,7 @@ func main() {
 
 		ec2Client := ec2.New(sess)
 
-		if nthConfig.EnablePrometheus {
+		if initMetricsErr == nil && nthConfig.EnablePrometheus {
 			go metrics.InitNodeMetrics(nthConfig, node, ec2Client)
 		}
 
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -120,25 +120,6 @@ func main() {
 		log.Fatal().Err(err).Msg("Unable to instantiate a node for various kubernetes node functions,")
 	}
 
-	cfg := aws.NewConfig().WithRegion(nthConfig.AWSRegion).WithEndpoint(nthConfig.AWSEndpoint).WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint)
-	sess := session.Must(session.NewSessionWithOptions(session.Options{
-		Config:            *cfg,
-		SharedConfigState: session.SharedConfigEnable,
-	}))
-	creds, err := sess.Config.Credentials.Get()
-	if err != nil {
-		log.Fatal().Err(err).Msg("Unable to get AWS credentials")
-	}
-	log.Debug().Msgf("AWS Credentials retrieved from provider: %s", creds.ProviderName)
-
-	ec2 := ec2.New(sess)
-
-	metrics, err := observability.InitMetrics(nthConfig, node, ec2)
-	if err != nil {
-		nthConfig.Print()
-		log.Fatal().Err(err).Msg("Unable to instantiate observability metrics,")
-	}
-
 	err = observability.InitProbes(nthConfig.EnableProbes, nthConfig.ProbesPort, nthConfig.ProbesEndpoint)
 	if err != nil {
 		nthConfig.Print()
@@ -165,6 +146,25 @@ func main() {
 	if nthConfig.AWSRegion == "" && nthConfig.EnableSQSTerminationDraining {
 		nthConfig.Print()
 		log.Fatal().Msgf("Unable to find the AWS region to process queue events.")
+	}
+
+	cfg := aws.NewConfig().WithRegion(nthConfig.AWSRegion).WithEndpoint(nthConfig.AWSEndpoint).WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint)
+	sess := session.Must(session.NewSessionWithOptions(session.Options{
+		Config:            *cfg,
+		SharedConfigState: session.SharedConfigEnable,
+	}))
+	creds, err := sess.Config.Credentials.Get()
+	if err != nil {
+		log.Fatal().Err(err).Msg("Unable to get AWS credentials")
+	}
+	log.Debug().Msgf("AWS Credentials retrieved from provider: %s", creds.ProviderName)
+
+	ec2Client := ec2.New(sess)
+
+	metrics, err := observability.InitMetrics(nthConfig, node, ec2Client)
+	if err != nil {
+		nthConfig.Print()
+		log.Fatal().Err(err).Msg("Unable to instantiate observability metrics")
 	}
 
 	recorder, err := observability.InitK8sEventRecorder(nthConfig.EmitKubernetesEvents, nthConfig.NodeName, nthConfig.EnableSQSTerminationDraining, nodeMetadata, nthConfig.KubernetesEventsExtraAnnotations, clientset)
@@ -226,7 +226,7 @@ func main() {
 			CancelChan:                    cancelChan,
 			SQS:                           sqsevent.GetSqsClient(sess),
 			ASG:                           autoscaling.New(sess),
-			EC2:                           ec2,
+			EC2:                           ec2Client,
 			BeforeCompleteLifecycleAction: func() { <-time.After(completeLifecycleActionDelay) },
 		}
 		monitoringFns[sqsEvents] = sqsMonitor
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -120,7 +120,7 @@ func main() {
 		log.Fatal().Err(err).Msg("Unable to instantiate a node for various kubernetes node functions,")
 	}
 
-	metrics, err := observability.InitMetrics(nthConfig)
+	metrics, err := observability.InitMetrics(nthConfig.EnablePrometheus, nthConfig.PrometheusPort)
 	if err != nil {
 		nthConfig.Print()
 		log.Fatal().Err(err).Msg("Unable to instantiate observability metrics,")
@@ -218,7 +218,7 @@ func main() {
 		ec2Client := ec2.New(sess)
 
 		if nthConfig.EnablePrometheus {
-			go metrics.InitNodeMetrics(node, ec2Client)
+			go metrics.InitNodeMetrics(nthConfig, node, ec2Client)
 		}
 
 		completeLifecycleActionDelay := time.Duration(nthConfig.CompleteLifecycleActionDelaySeconds) * time.Second
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -217,7 +217,9 @@ func main() {
 
 		ec2Client := ec2.New(sess)
 
-		go metrics.InitNodeMetrics(node, ec2Client)
+		if nthConfig.EnablePrometheus {
+			go metrics.InitNodeMetrics(node, ec2Client)
+		}
 
 		completeLifecycleActionDelay := time.Duration(nthConfig.CompleteLifecycleActionDelaySeconds) * time.Second
 		sqsMonitor := sqsevent.SQSMonitor{
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -120,6 +120,12 @@ func main() {
 		log.Fatal().Err(err).Msg("Unable to instantiate a node for various kubernetes node functions,")
 	}
 
+	metrics, err := observability.InitMetrics(nthConfig)
+	if err != nil {
+		nthConfig.Print()
+		log.Fatal().Err(err).Msg("Unable to instantiate observability metrics,")
+	}
+
 	err = observability.InitProbes(nthConfig.EnableProbes, nthConfig.ProbesPort, nthConfig.ProbesEndpoint)
 	if err != nil {
 		nthConfig.Print()
@@ -146,25 +152,6 @@ func main() {
 	if nthConfig.AWSRegion == "" && nthConfig.EnableSQSTerminationDraining {
 		nthConfig.Print()
 		log.Fatal().Msgf("Unable to find the AWS region to process queue events.")
-	}
-
-	cfg := aws.NewConfig().WithRegion(nthConfig.AWSRegion).WithEndpoint(nthConfig.AWSEndpoint).WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint)
-	sess := session.Must(session.NewSessionWithOptions(session.Options{
-		Config:            *cfg,
-		SharedConfigState: session.SharedConfigEnable,
-	}))
-	creds, err := sess.Config.Credentials.Get()
-	if err != nil {
-		log.Fatal().Err(err).Msg("Unable to get AWS credentials")
-	}
-	log.Debug().Msgf("AWS Credentials retrieved from provider: %s", creds.ProviderName)
-
-	ec2Client := ec2.New(sess)
-
-	metrics, err := observability.InitMetrics(nthConfig, node, ec2Client)
-	if err != nil {
-		nthConfig.Print()
-		log.Fatal().Err(err).Msg("Unable to instantiate observability metrics")
 	}
 
 	recorder, err := observability.InitK8sEventRecorder(nthConfig.EmitKubernetesEvents, nthConfig.NodeName, nthConfig.EnableSQSTerminationDraining, nodeMetadata, nthConfig.KubernetesEventsExtraAnnotations, clientset)
@@ -217,6 +204,21 @@ func main() {
 		}
 	}
 	if nthConfig.EnableSQSTerminationDraining {
+		cfg := aws.NewConfig().WithRegion(nthConfig.AWSRegion).WithEndpoint(nthConfig.AWSEndpoint).WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint)
+		sess := session.Must(session.NewSessionWithOptions(session.Options{
+			Config:            *cfg,
+			SharedConfigState: session.SharedConfigEnable,
+		}))
+		creds, err := sess.Config.Credentials.Get()
+		if err != nil {
+			log.Fatal().Err(err).Msg("Unable to get AWS credentials")
+		}
+		log.Debug().Msgf("AWS Credentials retrieved from provider: %s", creds.ProviderName)
+
+		ec2Client := ec2.New(sess)
+
+		go metrics.InitNodeMetrics(node, ec2Client)
+
 		completeLifecycleActionDelay := time.Duration(nthConfig.CompleteLifecycleActionDelaySeconds) * time.Second
 		sqsMonitor := sqsevent.SQSMonitor{
 			CheckIfManaged:                nthConfig.CheckTagBeforeDraining,
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -120,7 +120,20 @@ func main() {
 		log.Fatal().Err(err).Msg("Unable to instantiate a node for various kubernetes node functions,")
 	}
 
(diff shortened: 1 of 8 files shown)


phuhung273:
@stevehipwell can you please give me a help on this. Now i only stuck at writing unit test for `opentelemetry.go` since it doesn
t have any test. 

Do you think i should refactor this file along with this PR or should I open another separate PR (eg: make opentelemetry.go testable)

LikithaVemulapalli:
Yes I agree with @stevehipwell here, let's separate refactoring PR to have a clear idea on the changes made. @phuhung273 if you want to test your changes for this PR let me know I will approve and run so for the future commits you can verify if the existing tests are working, if there are any conflicts or not, once you change the PR status to ready I will run the workflow, appreciate for your contribution. Thanks!

tiationg-kho:
Hi @phuhung273, could you resolve the conflicts in `pkg/node/node_test.go`?

phuhung273:
Thanks and Happy new year @tiationg-kho. Conflicts resolved

pkg/ec2helper/ec2helper.go
```diff
@@ -0,0 +1,92 @@
...
+
+func New(ec2 ec2iface.EC2API) EC2Helper {
+	return EC2Helper{
+		ec2ServiceClient: ec2,
+	}
+}
+
+func (h EC2Helper) GetInstanceIdsByTagKey(tag string) ([]string, error) {
+	ids := []string{}
+	nextToken := ""
```
tiationg-kho, line 40: We could declare token without initialize the empty string as value here.

pkg/ec2helper/ec2helper.go
```diff
@@ -0,0 +1,92 @@
...
+
+	for {
+		result, err := h.ec2ServiceClient.DescribeInstances(&ec2.DescribeInstancesInput{
+			Filters: []*ec2.Filter{
```
tiationg-kho, line 44: Could we add a filter for instance state here?

pkg/ec2helper/ec2helper.go
```diff
@@ -0,0 +1,92 @@
...
+			},
+			NextToken: &nextToken,
+		})
+
+		if err != nil {
+			return ids, err
+		}
+
+		if result == nil || len(result.Reservations) == 0 ||
+			len(result.Reservations[0].Instances) == 0 {
```
tiationg-kho, line 58: We might have nil issue.

Maybe we do not need to treat empty result as error condition?

pkg/ec2helper/ec2helper.go
```diff
@@ -0,0 +1,92 @@
...
+			return ids, fmt.Errorf("failed to describe instances")
+		}
+
+		for _, reservation := range result.Reservations {
+			for _, instance := range reservation.Instances {
+				if instance.InstanceId == nil {
```
tiationg-kho, line 64: Might have nil issue here.

pkg/node/node.go
```diff
@@ -635,6 +635,32 @@ func (n Node) fetchKubernetesNode(nodeName string) (*corev1.Node, error) {
...
+	}
+
+	if matchingNodes == nil || len(matchingNodes.Items) == 0 {
+		return ids, nil
+	}
+
+	for _, node := range matchingNodes.Items {
+		// sample providerID: aws:///us-west-2a/i-0abcd1234efgh5678
+		parts := strings.Split(node.Spec.ProviderID, "/")
+		ids = append(ids, parts[len(parts)-1])
```
tiationg-kho, line 658: We could add some check here.

pkg/observability/opentelemetry.go
```diff
@@ -70,13 +82,46 @@ func InitMetrics(enabled bool, port int) (Metrics, error) {
 		return Metrics{}, fmt.Errorf("failed to start Go runtime metrics collection: %w", err)
 	}
 
-	if enabled {
-		serveMetrics(port)
+	if metrics.enabled {
+		metrics.initCronMetrics()
```
tiationg-kho, line 86: We should use a goroutine here.

pkg/observability/opentelemetry.go
```diff
@@ -70,13 +82,46 @@ func InitMetrics(enabled bool, port int) (Metrics, error) {
...
+	// Run a periodic task
+	ticker := time.NewTicker(30 * time.Second)
+	defer ticker.Stop()
+
+	for range ticker.C {
+		m.serveNodeMetrics()
+	}
+}
+
+func (m Metrics) serveNodeMetrics() {
```
tiationg-kho, line 103: We could add nil check for both results. (`instanceIdsMap`, `nodeInstanceIds`)

We should not use `instanceIdsMap` in second block if we got error from `GetInstanceIdsMapByTagKey`.

Consider we filter the nodes result based on instances result, we would not have a chance to record any result (nodes > instances). Do you think this is a potential issue?

cmd/node-termination-handler.go
```diff
@@ -120,7 +120,20 @@ func main() {
...
+		Config:            *cfg,
+		SharedConfigState: session.SharedConfigEnable,
+	}))
+	creds, err := sess.Config.Credentials.Get()
+	if err != nil {
+		log.Fatal().Err(err).Msg("Unable to get AWS credentials")
+	}
+	log.Debug().Msgf("AWS Credentials retrieved from provider: %s", creds.ProviderName)
+
+	ec2 := ec2.New(sess)
```
tiationg-kho, line 134: We should not use `ec2` as var name. This will cause issue in later like init `sqsevent.SQSMonitor`.

cmd/node-termination-handler.go
```diff
@@ -120,7 +120,20 @@ func main() {
 		log.Fatal().Err(err).Msg("Unable to instantiate a node for various kubernetes node functions,")
 	}
 
-	metrics, err := observability.InitMetrics(nthConfig.EnablePrometheus, nthConfig.PrometheusPort)
+	cfg := aws.NewConfig().WithRegion(nthConfig.AWSRegion).WithEndpoint(nthConfig.AWSEndpoint).WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint)
```
tiationg-kho, line 123: We have some code to setup aws related config like below. So we should create client after that.
```
if nthConfig.AWSRegion == "" && nodeMetadata.Region != "" {
    nthConfig.AWSRegion = nodeMetadata.Region
} else if nthConfig.AWSRegion == "" && nthConfig.QueueURL != "" {
    nthConfig.AWSRegion = getRegionFromQueueURL(nthConfig.QueueURL)
    log.Debug().Str("Retrieved AWS region from queue-url: \"%s\"", nthConfig.AWSRegion)
}
if nthConfig.AWSRegion == "" && nthConfig.EnableSQSTerminationDraining {
    nthConfig.Print()
    log.Fatal().Msgf("Unable to find the AWS region to process queue events.")
}
```

tiationg-kho (commented):
Hi @phuhung273,

Thanks for resolving the conflicts. I have left some comments. 
Also, we can run the e2e test in local (docker desktop) to make sure our modification is valid.
- Run all e2e test cases: `make e2e-test`
- Run certain e2e test case: `./test/k8s-local-cluster-test/run-test -a ./test/e2e/<test-case> -d -b e2e-test`

pkg/ec2helper/ec2helper.go
phuhung273: Of course we can, but I wonder if we should ? I saw cases where instances enter `Stopped` state instead of `Terminated` . Without filterring user can discover such case.

pkg/observability/opentelemetry.go
phuhung273: Thanks, I have addressed the 2 nil check.

About the edge cases, I did think of if but could not find anyway to do 2 metrics separately:
- For instances we already have a NTH managed tag so we can filter base on it
- For nodes i cannot see any label/annotation that we can filter upon

Therefore, I decided to filter nodes based on instances result. Do you know any info we can use to filter nodes independently

pkg/ec2helper/ec2helper.go
phuhung273: Thanks, this is fixed now

pkg/ec2helper/ec2helper.go
phuhung273: Thanks, just added nil check and removed empty array check

pkg/ec2helper/ec2helper.go
phuhung273: Thanks, nil check added

pkg/node/node.go
phuhung273: Thanks, I have added string split check and 'i-xxxxx' check

pkg/observability/opentelemetry.go
phuhung273: Yes, modified to use goroutine

cmd/node-termination-handler.go
phuhung273: var name changed to ec2Client

cmd/node-termination-handler.go
phuhung273: Thanks, i have moved the metrics initialization after AWS config setup

pkg/ec2helper/ec2helper.go
tiationg-kho: Thanks for addressing the comments.

Maybe we could rename `nodesGauge` and `instancesGauge` to let these 2 metrics more self-explain?

Then we could decide we need a filter here or not. And would also know should separate these 2 metrics, or filter one based on another (`opentelemetry.go`).

phuhung273: Thanks, that makes sense. I have renamed to `nthTaggedNodesGauge` and `nthTaggedInstancesGauge`

phuhung273:
Thanks @tiationg-kho for your instruction on the e2e test, took a while to realize where does it fail. I will try to include the 2 new metrics in e2e.

cmd/node-termination-handler.go
```diff
@@ -215,6 +215,10 @@ func main() {
 		}
 		log.Debug().Msgf("AWS Credentials retrieved from provider: %s", creds.ProviderName)
 
+		ec2Client := ec2.New(sess)
+
+		go metrics.InitNodeMetrics(node, ec2Client)
```
tiationg-kho, line 220: We can check `EnablePrometheus` before calling.

phuhung273: Thanks, `EnablePrometheus` check added

phuhung273:
e2e metrics test in SQS mode added, phew... Please let me know your thought @tiationg-kho, thanks

cmd/node-termination-handler.go
```diff
@@ -215,6 +215,12 @@ func main() {
 		}
 		log.Debug().Msgf("AWS Credentials retrieved from provider: %s", creds.ProviderName)
 
+		ec2Client := ec2.New(sess)
+
+		if nthConfig.EnablePrometheus {
```
tiationg-kho, line 220: Could we add nil check for `metrics` here in case we have any error during `observability.InitMetrics`?

test/e2e/prometheus-metrics-sqs-test
```diff
@@ -0,0 +1,241 @@
...
+    fi
+
+    if [[ $cordoned -eq 1 && $(kubectl get deployments regular-pod-test -o=jsonpath='{.status.unavailableReplicas}') -eq 1 ]]; then
+        echo "✅ Verified the regular-pod-test pod was evicted!"
+        evicted=1
+    fi
+
+    if [[ ${evicted} -eq 1 && $(kubectl exec -i "${localstack_pod}" -- bash -c "$GET_ATTRS_SQS_CMD" | jq '(.Attributes.ApproximateNumberOfMessagesNotVisible|tonumber) + (.Attributes.ApproximateNumberOfMessages|tonumber)' ) -eq 0 ]]; then
+        kubectl exec -i "${localstack_pod}" -- bash -c "$GET_ATTRS_SQS_CMD"
+        echo "✅ Verified the message was deleted from the queue after processing!"
```
tiationg-kho, line 170: We could add a `break` here.

test/e2e/prometheus-metrics-sqs-test
```diff
@@ -0,0 +1,241 @@
...
+    query="$metric_name{$labels}"
+    counter_value=$(echo "$METRICS_RESPONSE" | grep -E "${query}[[:space:]]+[0-9]+" | awk '{print $NF}')
+    if (($counter_value < 1)); then
+        echo "❌ Failed counter count for metric action:$action"
+        exit 5
+    fi
+    echo "✅ Fetched counter:$counter_value for metric with action:$action"
+done
+
+for gauge in nth_tagged_instances; do
```
tiationg-kho, line 230: This cannot pass the shell check (`make shellcheck`).

Could we remove the for-loop and just declare the gauge var like (`gauge="nth_tagged_instances"`)

tiationg-kho (commented):
Thank you for adding a new e2e test case! Just left few comments.

@LikithaVemulapalli could you please take a look too?

test/e2e/prometheus-metrics-sqs-test
```diff
@@ -0,0 +1,241 @@
...
+    fi
+
+    echo "Assertion Loop $i/$TAINT_CHECK_CYCLES, sleeping for $TAINT_CHECK_SLEEP seconds"
+    sleep $TAINT_CHECK_SLEEP
+done
+
+if [[ $cordoned -eq 0 ]]; then
+    echo "❌ Worker node was not cordoned"
+else
+    echo "❌ regular-pod-test was not evicted"
```
tiationg-kho, line 180: This check is not quite correct in the context of this test case.

LikithaVemulapalli (changes requested):
left some inline comments, thanks for your contribution :)

cmd/node-termination-handler.go
phuhung273: I tried to, but the compiler doesn't like it

pkg/observability/opentelemetry.go
```diff
@@ -70,13 +80,49 @@ func InitMetrics(enabled bool, port int) (Metrics, error) {
...
+	instanceIdsMap, err := m.ec2Helper.GetInstanceIdsMapByTagKey(m.nthConfig.ManagedTag)
+	if err != nil || instanceIdsMap == nil {
+		log.Err(err).Msg("Failed to get AWS instance ids")
+		return
+	} else {
+		m.InstancesRecord(int64(len(instanceIdsMap)))
```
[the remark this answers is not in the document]
phuhung273: Thank you for spotting this. I should have been more consistent.

test/e2e/prometheus-metrics-sqs-test
phuhung273: Perfect, appreciate your careful review

test/e2e/prometheus-metrics-sqs-test
phuhung273: At first, I just tried to test the 2 new metrics. Later found out that https://github.com/aws/aws-node-termination-handler/blob/main/test/e2e/prometheus-metrics-test only test the metrics in IMDS mode.

So I decided to also include other metrics as well. Thats why the file name is `prometheus-metrics-sqs`: test all metrics in SQS mode. How do you think ?

pkg/ec2helper/ec2helper.go
```diff
@@ -0,0 +1,98 @@
...
+		if result.NextToken == nil {
+			break
+		}
+		nextToken = *result.NextToken
+	}
+
+	return ids, nil
+}
+
+func (h EC2Helper) GetInstanceIdsMapByTagKey(tag string) (map[string]bool, error) {
```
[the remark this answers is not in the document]
phuhung273: If I understand correctly, querying DescribeAPI with `tag-key`  only care about the Tag's key, tag's value doesn't matter  https://github.com/aws/aws-node-termination-handler/blob/316906ea1bd59100d1f749cc935253fe65dc1107/pkg/ec2helper/ec2helper.go#L44-L49

pkg/ec2helper/ec2helper.go
```diff
@@ -0,0 +1,98 @@
...
+	idMap := map[string]bool{}
+	ids, err := h.GetInstanceIdsByTagKey(tag)
+	if err != nil {
+		return nil, err
+	}
+
+	if ids == nil {
```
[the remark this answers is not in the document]
phuhung273: Thank you, error message modified. Hope it is clear enough.

pkg/ec2helper/ec2helper_test.go
```diff
@@ -0,0 +1,74 @@
...
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package ec2helper_test
```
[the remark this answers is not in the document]
phuhung273: My mistake, added 6 more cases here. Please let me know if any more cases should be added.

pkg/node/node.go
```diff
@@ -635,6 +637,41 @@ func (n Node) fetchKubernetesNode(nodeName string) (*corev1.Node, error) {
...
+		// sample providerID: aws:///us-west-2a/i-0abcd1234efgh5678
+		parts := strings.Split(node.Spec.ProviderID, "/")
+		if len(parts) < 2 {
+			log.Warn().Msgf("Found invalid providerID: %s", node.Spec.ProviderID)
+			continue
+		}
+
+		instanceId := parts[len(parts)-1]
+		if instanceIDRegex.MatchString(instanceId) {
+			ids = append(ids, parts[len(parts)-1])
```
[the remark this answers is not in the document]
phuhung273: Thank you, added more info to warning message.

pkg/node/node_test.go
```diff
@@ -379,6 +384,31 @@ func TestUncordonIfRebootedTimeParseFailure(t *testing.T) {
 	h.Assert(t, err != nil, "Failed to return error on UncordonIfReboted failure to parse time")
 }
 
+func TestFetchKubernetesNodeInstanceIds(t *testing.T) {
```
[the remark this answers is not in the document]
phuhung273: Added 2 more test cases, invalid ProviderId case contains 4 invalid samples inside.

test/e2e/prometheus-metrics-sqs-test
phuhung273: Thank you, will also run shellcheck from now on.

pkg/node/node.go
```diff
@@ -635,6 +637,41 @@ func (n Node) fetchKubernetesNode(nodeName string) (*corev1.Node, error) {
...
+
+	if matchingNodes == nil || matchingNodes.Items == nil {
+		return nil, fmt.Errorf("failed to list nodes")
+	}
+
+	for _, node := range matchingNodes.Items {
+		// sample providerID: aws:///us-west-2a/i-0abcd1234efgh5678
+		parts := strings.Split(node.Spec.ProviderID, "/")
+		if len(parts) < 2 {
+			log.Warn().Msgf("Found invalid providerID: %s", node.Spec.ProviderID)
```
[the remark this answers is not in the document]
phuhung273: Yes sir, warning message updated. No more dummy error message from now on.

test/e2e/prometheus-metrics-sqs-test
tiationg-kho: Maybe we could remove this if-else check, and replace by adding a status variable before `TAINT_CHECK_CYCLES` for-loop, and check it after the for-loop to prevent the misleading scenario here (even we already cordoned and evicted but still get a "not evicted" in our log)?

The metric part is good.

cmd/node-termination-handler.go
tiationg-kho: Sorry, "nil check" is incorrect. Since we return a zero-value struct when err exists.

Could we rename the error then check that error here?
```
metrics, initMetricsErr := observability.InitMetrics(nthConfig.EnablePrometheus, nthConfig.PrometheusPort)
if initMetricsErr != nil {
    nthConfig.Print()
    log.Fatal().Err(initMetricsErr).Msg("Unable to instantiate observability metrics,")
}
```

phuhung273: That make sense, understand your idea now. Added `initMetricsErr` nil check. Thank you

test/e2e/prometheus-metrics-sqs-test
phuhung273: Oh must be my copy-paste fault :( Removed but I dont understand the status check youre mentioning. Also later part cover the metrics, maybe thats enough ?

phuhung273: Thanks for your detailed instruction. Added check for `cordoned`, `evicted`, and `message_deleted`.

tiationg-kho (approved):
lgtm

LikithaVemulapalli (approved):
Looks good to me. Appreciate the efforts @phuhung273

LikithaVemulapalli:
please do not merge change until all the tests are successful. Thanks

phuhung273:
Thanks for your kind guidance @tiationg-kho @LikithaVemulapalli

PR #1094, opened 2024-12-04
Fix e2e tests failure in action of build and test

tiationg-kho:
**Issue #, if available:**

**Description of changes:**
- Add custom service client when perform e2e test
- Make sure AEMM is up before NTH
- Fix validation in prometheus-metrics-test

### DIFF
--- a/pkg/ec2metadata/ec2metadata.go
+++ b/pkg/ec2metadata/ec2metadata.go
@@ -122,18 +122,6 @@ type NodeMetadata struct {
 
 // New constructs an instance of the Service client
 func New(metadataURL string, tries int) *Service {
-	if strings.HasPrefix(metadataURL, "http://amazon-ec2-metadata-mock-service.default.svc.cluster.local") {
-		return &Service{
-			metadataURL: metadataURL,
-			tries:       tries,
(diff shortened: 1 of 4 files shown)


tiationg-kho:
This pattern follows the approach in webhook-http-proxy-test, and webhook-secret-test. 

I think we should be good for only updating these 3 tests (emit-events-test, prometheus-metrics-test, and webhook-test).

pkg/ec2metadata/ec2metadata.go
```diff
@@ -122,6 +122,18 @@ type NodeMetadata struct {
 
 // New constructs an instance of the Service client
 func New(metadataURL string, tries int) *Service {
+	if strings.HasPrefix(metadataURL, "http://amazon-ec2-metadata-mock-service.default.svc.cluster.local") {
```
LikithaVemulapalli, line 125: add a comment to clarify it here or either remove it if tests are passing successfully without it

tiationg-kho: Fixed by removing.

LikithaVemulapalli (approved):
/lgtm

Issue #1024, opened 2024-06-13
ASG launch lifecycle events showing up as `UnknownInterruption`


stevehipwell:
**Describe the bug**
All events for the ASG launch lifecycle are categorised as `UnknownInterruption` instead of the expected `ASGLifecycle`.

**Steps to reproduce**
Use the ASG launch functionality with K8s events enabled.

**Expected outcome**
I'd expect to see the events categorised as `ASGLifecycle`.

**Application Logs**
n/a

**Environment**

* NTH App Version: 1.22.0
* NTH Mode (IMDS/Queue processor): Queue processor
* OS/Arch:
* Kubernetes version: 
* Installation method: Helm

stevehipwell:
CC @LikithaVemulapalli

stevehipwell:
/not-stale

CC @LikithaVemulapalli

LikithaVemulapalli:
Hello @stevehipwell, we will further see that this issue is not marked as stale and will give an update on this as soon as possible. Thanks!

PR #1096, opened 2024-12-05
Fix ASGLifecycle showing up as UnknownInterruption on K8 events (#1024)

hyeong01:
**Issue #, if available:**
#1024 

**Description of changes:**
getReasonForKindV1 used for K8sEventsRecorder was not catching `ASGLifeCycle` event, making the event fall to the default case of `unknownReason`. Added a new case that catches this `ASGLifeCycle` event. 

### DIFF
--- a/pkg/observability/k8s-events.go
+++ b/pkg/observability/k8s-events.go
@@ -169,6 +169,8 @@ func getReasonForKindV1(eventKind, monitorKind string) string {
 		return spotITNReason
 	case monitor.RebalanceRecommendationKind:
 		return rebalanceRecommendationReason
+	case monitor.ASGLifecycleKind:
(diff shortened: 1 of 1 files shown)


Lu-David:
Can you briefly add description on how you tested it? Include any screenshots with non-sensitive information would be helpful. I will also update the template to have a testing section.

hyeong01:
Hi David, what do you think about also tapping on a related issue? StateChange is appearing as UnknownInterruption as well. This can be fixed by adding one more case loop for StateChangeKind. Seems like a good time to fix and test it together.

Lu-David:
Good catch, we should fix that as well. Thanks!

Lu-David (approved):

Issue #717, opened 2022-11-02
No SQS retry on "read: connection reset by peer"


eugenea:
**Describe the bug**
NTH does not retry request over AWS SDK API to retrieve SQS queue message.

**Steps to reproduce**
Close firewall to SQS AWS endpoint and try to monitor for SQS events.

**Expected outcome**
The network layer cannot be guaranteed to be reliable so need to implement retry logic here.

**Application Logs**
```
WRN There was a problem monitoring for events error="RequestError: send request failed\ncaused by: Post \"https://sqs.us-west-2.amazonaws.com/\": read tcp 100.100.xx.xx:xxxx->10.xx.xx.xx:443: read: connection reset by peer" event_type=SQS_TERMINATE
```

**Environment**

* NTH App Version: v1.16.3
* NTH Mode (IMDS/Queue processor): Queue processor 
* OS/Arch: Linux
* Kubernetes version: v1.21.14-eks 
* Installation method: deployment

The check that denies retry is [here](https://github.com/aws/aws-sdk-go/blob/main/aws/request/connection_reset_error.go)
For V1 of AWS SDK the fix should be custom retryer which re-implements [should retry](https://github.com/aws/aws-sdk-go/blob/e2d6cb448883e4f4fcc5246650f89bde349041ec/aws/client/default_retryer.go#L131), and custom retryer should be injected [here](https://github.com/aws/aws-node-termination-handler/blob/main/cmd/node-termination-handler.go#L189),  however upgrade to V2 of AWS SKD should fix this issue automatically,  because it does not make distinction between different kinds of [connection reset](https://github.com/aws/aws-sdk-go-v2/blob/b7d8e15425d2f86a0596e8d7db2e33bf382a21dd/aws/retry/retryable_error.go#L103) and retries them all which is desired behavior here.

snay2:
Thank you for the suggestion! Upon first reading, we would favor doing the upgrade to v2 of the AWS SDK if it can handle this logic automatically.

eugenea:
Do you have any timeline/plan for v2 upgrade?

jillmon:
@eugenea, the [beta version](https://github.com/aws/aws-node-termination-handler/tree/v2) of the NTH v2 upgrade has recently been released. Have you had a chance to investigate whether then new SDK can handle this use case?

tiationg-kho:
Added a custom retryer for SQS in NTH v1. Please let us know if there is any other issue. Thank you!

[...]

PR #1099, opened 2024-12-11
Add sqs client retryer

tiationg-kho:
**Issue #, if available:**
#717

**Description of changes:**
- Add sqs client retryer
- Add sqs client retryer unit test

### DIFF
--- a/cmd/node-termination-handler.go
+++ b/cmd/node-termination-handler.go
@@ -16,7 +16,6 @@ package main
 import (
 	"context"
 	"fmt"
-	"github.com/aws/aws-node-termination-handler/pkg/monitor/asglifecycle"
 	"os"
 	"os/signal"
 	"strings"
@@ -31,6 +30,7 @@ import (
 	"github.com/aws/aws-node-termination-handler/pkg/interruptioneventstore"
 	"github.com/aws/aws-node-termination-handler/pkg/logging"
 	"github.com/aws/aws-node-termination-handler/pkg/monitor"
+	"github.com/aws/aws-node-termination-handler/pkg/monitor/asglifecycle"
 	"github.com/aws/aws-node-termination-handler/pkg/monitor/rebalancerecommendation"
 	"github.com/aws/aws-node-termination-handler/pkg/monitor/scheduledevent"
 	"github.com/aws/aws-node-termination-handler/pkg/monitor/spotitn"
@@ -43,7 +43,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/ec2"
-	"github.com/aws/aws-sdk-go/service/sqs"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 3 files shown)


cmd/node-termination-handler.go
```diff
@@ -223,7 +222,7 @@ func main() {
 			QueueURL:                      nthConfig.QueueURL,
 			InterruptionChan:              interruptionChan,
 			CancelChan:                    cancelChan,
-			SQS:                           sqs.New(sess),
+			SQS:                           sqsevent.GetSqsClient(sess),
 			ASG:                           autoscaling.New(sess),
```
Lu-David, line 226: should we also add retry for asg client?

pkg/monitor/sqsevent/sqs-retryer.go
```diff
@@ -0,0 +1,48 @@
...
+		(req.Error != nil && strings.Contains(req.Error.Error(), "connection reset"))
+}
+
+func GetSqsClient(sess *session.Session) *sqs.SQS {
+	return sqs.New(sess, &aws.Config{
+		Retryer: SqsRetryer{
+			DefaultRetryer: client.DefaultRetryer{
+				NumMaxRetries:    client.DefaultRetryerMaxNumRetries,
+				MinRetryDelay:    client.DefaultRetryerMinRetryDelay,
+				MaxRetryDelay:    1200 * time.Millisecond,
```
Lu-David, line 42: any particular reason why we don't set MaxRetryDelay to be DefaultRetryerMaxRetryDelay and same for DefaultRetryerMinThrottleDelay? https://github.com/aws/aws-sdk-go/blob/7112c0a0c2d01713a9db2d57f0e5722225baf5b5/aws/client/default_retryer.go#L49C2-L49C29

pkg/monitor/sqsevent/sqs-retryer.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/aws/aws-sdk-go/aws/request"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/sqs"
+)
+
+type SqsRetryer struct {
+	client.DefaultRetryer
+}
+
+func (r SqsRetryer) ShouldRetry(req *request.Request) bool {
```
Lu-David, line 31: nit: personally I would prefer more descriptive variable names (perhaps retryer instead or r), just so I don't have to search for the type.

pkg/monitor/sqsevent/sqs-retryer_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/aws/aws-sdk-go/aws/request"
+	"github.com/aws/aws-sdk-go/aws/session"
+)
+
+type temporaryError struct {
+	error
+	temp bool
+}
+
+func TestGetSqsClient(t *testing.T) {
```
Lu-David, line 36: this test case seems unnecessary because it's testing an implementation detail rather than behavior.

pkg/monitor/sqsevent/sqs-retryer.go
tiationg-kho: Fixed by adding doc comment.

cmd/node-termination-handler.go
tiationg-kho: We could keep use default retryer for asg client.

pkg/monitor/sqsevent/sqs-retryer_test.go
tiationg-kho: Would keep test case to test custom retryer config is consumed or not.

Lu-David (approved):
lgtm

PR #1107, opened 2024-12-17
Deprecate old k8s version

tiationg-kho:
**Issue #, if available:**

**Description of changes:**
- Deprecate old k8s version

Kubernetes Version: 1.31

Test in local by `make compatibility-test`

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -191,7 +191,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        k8sVersion: ["1.23", "1.24", "1.25", "1.26", "1.27", "1.28", "1.29", "1.30"]
+        k8sVersion: ["1.29", "1.30", "1.31"]
     steps:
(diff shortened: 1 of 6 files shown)


LikithaVemulapalli (approved):
/lgtm

Lu-David:
Any particular reason why we are deprecating these particular versions? Seems like version 1.27 and 1.28 came out in 2023, which is still pretty recent.

LikithaVemulapalli (commented):
From the official [doc](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html), it seems only 1.23 is deprecated, we usually only remove deprecated versions. Also let's remove 1.24 as it will be deprecated January 31, 2025

LikithaVemulapalli (changes requested):
reverting the approval until the versions are fixed.

LikithaVemulapalli (approved):
we can create new clusters only from k8s version 1.29 using aws eks service, keep the original implementation to remove earlier versions

Issue #1124, opened 2025-01-28
Multi-Attach error for volume "pvc-xxxx". Volume is already exclusively attached to one node and can't be attached


JnMik:
Hey guys ! 

**Describe the feature**

We've all probably ever see this error message, when an ASG is refreshing instances, but the pod is not able to respawn on the new instance because it's volume seems to be stuck in a state attached to the previous machine. Well, this can be fixed by manually finding the PVCID and deleting the kubernetes volumeattachments corresponding, or simply wait for a couple minutes that feels like forever. 

I was thinking, wouldn't it be nice if aws-node-termination-handler handles that ? Or is it supposed to be handling it already ? 

Or is there another automated way that I am not aware of ? 

Cheers

PR #1132, opened 2025-02-21
Add enable-out-of-service-taint flag

tiationg-kho:
**Issue #, if available:**
#1124

**Description of changes:**

Add enable-out-of-service-taint flag
- Enabling this feature will add out-of-service taint to node after cordon/drain process which would forcefully evict pods without matching tolerations and detach persistent volumes
- This could prevent the PVC multi-attach error

Kubernetes Version: 1.31

Test with queue mode NTH: refreshing ASG and monitoring the k8s event

Ref:
[PVC attaching takes much time](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1302)
[Kubernetes 1.28: Non-Graceful Node Shutdown Moves to GA](https://kubernetes.io/blog/2023/08/16/kubernetes-1-28-non-graceful-node-shutdown-ga/)


Lu-David:
Were you able to test this on an actual ASG perchance? I wonder if it would be simple enough to reproduce the issue with start-instance-refresh for an ASG or using EKS? And also test that this fix will get rid of that multi-attach error?

pkg/config/config.go
```diff
@@ -215,6 +218,7 @@ func ParseCliArgs() (config Config, err error) {
 	flag.BoolVar(&config.CordonOnly, "cordon-only", getBoolEnv(cordonOnly, false), "If true, nodes will be cordoned but not drained when an interruption event occurs.")
 	flag.BoolVar(&config.TaintNode, "taint-node", getBoolEnv(taintNode, false), "If true, nodes will be tainted when an interruption event occurs.")
 	flag.StringVar(&config.TaintEffect, "taint-effect", getEnv(taintEffect, taintEffectDefault), "Sets the effect when a node is tainted.")
+	flag.BoolVar(&config.EnableOutOfServiceTaint, "enable-out-of-service-taint", getBoolEnv(enableOutOfServiceTaintConfigKey, enableOutOfServiceTaintDefault), "If ture, nodes will be tainted as out-of-service after we cordon/drain the nodes when an interruption event occurs.")
```
Lu-David, line 221: typo: If true

LikithaVemulapalli (dismissed):
/lgtm, you can merge it after the typo fix. Thanks

tiationg-kho:
> Were you able to test this on an actual ASG perchance? I wonder if it would be simple enough to reproduce the issue with start-instance-refresh for an ASG or using EKS? And also test that this fix will get rid of that multi-attach error?

Have tested by deploying StatefulSet with Volume. First, we recreated the PVC multi-attach error through ASG refresh. 

Then triggered ASG refresh multiple times (let NTH enable the out-of-service taint flag). During each round, we monitored and confirmed that the error did not occur.

pkg/config/config.go
tiationg-kho: Fixed.

PR #1142, opened 2025-03-12
Modified cache and ubuntu versions to resolve the build failures

LikithaVemulapalli:
**Issue #, if available:**
We are using deprecated versions in our workflows causing Github Actions workflow failures.

**Description of changes:**
- Modified the `actions/cache` versions from v2 to v4 as per recommendation in official notice.
- Modified `ubuntu` versions as well in workflows as the current versions will be deprecated by March end.


tiationg-kho (approved):
lgtm

Issue #1148, opened 2025-03-25
Propagate log levels from k8s dependencies


der-eismann:
**Describe the bug**
When looking at the logs I can see that all messages directly from NTH have a level like `INF` or `WRN`. However if the message is coming from a k8s dependency (I think), the log level is given with `???` or omitted for JSON logging. It would be great if we could have a log level for these messages as well.

**Steps to reproduce**
1. Run NTH and drain a node

**Expected outcome**
There should be no `???` for log levels anymore

**Application Logs**
```
2025/03/25 08:07:27 INF Draining the node
2025/03/25 08:07:27 ??? WARNING: ignoring DaemonSet-managed Pods: default/abc-wlm7t, default/def-rl86f, default/ghi-bb65s, …
2025/03/25 08:07:27 ??? evicting pod default/my-pod-79dc9db6d5-g9rln
```

**Environment**

* NTH App Version: v1.25.0
* NTH Mode (IMDS/Queue processor): Queue
* OS/Arch: Container / Bottlerocket arm64
* Kubernetes version: 1.32
* Installation method: Helm

der-eismann:
Solution may be the same as for #1150 with the [SetLogger function in klog](https://pkg.go.dev/k8s.io/klog/v2@v2.130.1#SetLogger)

der-eismann:
Turns out that this doesn't help as this comes from kubectl and that's using plain `fmt` and not `klog` 😕

tiationg-kho:
We have included fix in the latest release. Thanks

PR #1153, opened 2025-04-03
Fix log level in drain helper

tiationg-kho:
**Issue #, if available:**
#1148

**Description of changes:**

Fix log level in drain helper by passing custom loggers.

Ref:
- `evictPods` and `fmt.Fprintf` in [pkg/drain/drain.go](https://github.com/kubernetes/kubectl/blob/fec9d5b3d5b1c3addf8cebd3c2437257ea5e4332/pkg/drain/drain.go#L268)
- `Fprintf` in [src/fmt/print.go](https://github.com/golang/go/blob/master/src/fmt/print.go#L222-L225)

Kubernetes Version: 1.32

Test by:
- e2e test
- deploy NTH by private ECR image; trigger spot interruption; validate the log messages

Prev:
```
INF Draining the node
??? evicting pod kube-system/ebs-csi-controller-xxx
??? error when evicting pods/"ebs-csi-controller-xxx" -n "kube-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
```
Cur:
```
INF Draining the node
INF evicting pod kube-system/ebs-csi-controller-xxx
ERR error when evicting pods/"ebs-csi-controller-xxx" -n "kube-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
```

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -89,6 +89,18 @@ type Node struct {
 	uptime      uptime.UptimeFuncType
 }
 
+type ZerologWriter struct {
+	logger zerolog.Logger
+}
+
+type StdWriter struct {
+	*ZerologWriter
+}
+
+type ErrWriter struct {
+	*ZerologWriter
+}
+
 // New will construct a node struct to perform various node function through the kubernetes api server
 func New(nthConfig config.Config, clientset *kubernetes.Clientset) (*Node, error) {
 	drainHelper, err := getDrainHelper(nthConfig, clientset)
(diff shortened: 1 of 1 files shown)


pkg/node/node.go
Lu-David, line 1: I see some of our test files are using log.Logger instead of ZeroLog.Logger. Do they need to be updated as well?

Lu-David: non-blocking btw

Lu-David (approved):

pkg/node/node.go
tiationg-kho: Have checked they are using same package. For the test cases, they do not validate base on log level. Might just keep them.

Issue #1277, opened 2026-05-20
addTaint uses full-object Update causing silent taint failures at high concurrency


mcornea:
**Describe the bug**
When NTH is configured with a high WORKERS value (>=30) and processes a large batch of concurrent spot interruption events, approximately 6-13% of taint operations silently fail. The addTaint function uses a full object Nodes().Update() that conflicts with concurrent cordon operations due to resourceVersion mismatch (HTTP 409), and the 5-second retry budget with 750ms intervals is insufficient to recover under high concurrency. Additionally, the PreDrainTask closure in spot-itn-event.go swallows the taint error (return nil instead of return err), so failed taints are never retried and the SQS message is deleted as if processing succeeded.

**Steps to reproduce**
1. Deploy NTH in Queue Processor mode with WORKERS=50 and TAINT_NODE=true
2. Have 100 spot instances managed by NTH
3. Inject 100 spot interruption SQS messages simultaneously (or trigger 100 concurrent spot interruptions)
4. Observe NTH logs: all 100 nodes show "Node successfully cordoned and drained" but only 87-91 show "Successfully added taint on node"
5. Repeat with WORKERS=10: all 100 taints succeed

**Expected outcome**
All 100 nodes should be tainted regardless of WORKERS value. If a taint operation fails, the error should be propagated so the SQS message is not deleted and the event is retried on the next poll cycle.

**Application Logs**
For a successful node, the log shows both taint and cordon/drain:
Successfully added taint on node  taint_key=aws-node-termination-handler/spot-itn node_name=ip-10-0-79-63.us-east-2.compute.internal Node successfully cordoned and drained  node_name=ip-10-0-79-63.us-east-2.compute.internal

For a failed node, only the cordon/drain message appears - the taint error is logged at error level but processing continues as if it succeeded:
Error while adding taint on node  taint_key=aws-node-termination-handler/spot-itn node_name=ip-10-0-79-156.us-east-2.compute.internal error="Operation cannot be fulfilled on nodes \"ip-10-0-79-156.us-east2.compute.internal\": the object has been modified; please apply your changes to the latest version and try again" Unable to taint node with taint aws-node-termination-handler/spot-itn:spot-itn-event-... Node successfully cordoned and drained  node_name=ip-10-0-79-156.us-east-2.compute.internal

**Environment**

- NTH App Version: latest (main branch as of 2026-05-18)
- NTH Mode (IMDS/Queue processor): Queue Processor
[...]

PR #1278, opened 2026-05-20
Fix silent taint loss under concurrent spot interruptions

mcornea:
**Issue #, if available:**
https://github.com/aws/aws-node-termination-handler/issues/1277

**Description of changes:**
addTaint uses a full-object Nodes().Update() that conflicts with concurrent cordon operations due to resourceVersion mismatch (HTTP 409 Conflict). Under high WORKERS values (>=30), the 5-second retry budget with 750ms fixed intervals is exhausted for ~6-13% of taint attempts. Additionally, the PreDrainTask closure in spot-itn-event.go swallows the taint error (returns nil), so the caller sees success, processing continues to cordon and drain, and the SQS message is deleted. No retry ever occurs for the failed taint.

This fix:
- Replaces Nodes().Update() with Nodes().Patch() using StrategicMergePatchType so taint patches don't conflict with concurrent cordon patches (they target different fields: spec.taints vs spec.unschedulable)
- Always fetches fresh node state before building the patch (removes the wasted first retry on a stale DeepCopy)
- Increases retry budget from 5s/750ms to 15s/500ms
- Propagates the taint error from PreDrainTask so the SQS message is not deleted and the event is retried on the next poll cycle

Environment (Linux / Windows): Linux (RHEL CoreOS 9.8)
Kubernetes Version: v1.35.4 (OpenShift 4.22)

Tested on a ROSA HCP cluster with 100 spot instances (c5a.xlarge). Injected 100 simultaneous synthetic SQS spot interruption messages and verified that all 100 nodes were tainted at WORKERS=50. Before the fix, only 87-91 out of 100 taints succeeded at that concurrency level.

### DIFF
--- a/pkg/monitor/sqsevent/spot-itn-event.go
+++ b/pkg/monitor/sqsevent/spot-itn-event.go
@@ -96,7 +96,7 @@ func (m SQSMonitor) spotITNTerminationToInterruptionEvent(event *EventBridgeEven
 		if err != nil {
 			log.Err(err).Msgf("Unable to taint node with taint %s:%s", node.SpotInterruptionTaint, interruptionEvent.EventID)
 		}
-		return nil
+		return err
 	}
 	return &interruptionEvent, nil
 }
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -77,8 +77,8 @@ const (
 )
 
 var (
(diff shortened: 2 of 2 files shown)


enxebre:
thanks! lgtm

hyeong01:
Closed as described in: #1277. Thank you @mcornea!

PR #1279, opened 2026-06-08
Resolve concurrency problem causing skipping taint by surfacing error and retrying with SQS

hyeong01:
**Issue #, if available:**
#1277

**Description of changes:**
As mentioned in the issue above, concurrency problems leads to taint errors. In specific, for situations over 100+ nodes getting de-provisioned simultaneously, 3-6% of nodes fails to be tainted and proceeds to be cordoned. This was due to the taint error not surfacing, and the message disappearing from the queue even without a successful taint. The new change explicitly returns the taint error and fails the pre-drain task so that the failed message can be handled by SQS retry logic.

Also added `DeleteSqsMsgIfNodeNotFound` condition (passed in as config param) in case the user wants to delete the message if the node is gone, to maintain consistent behavior with postDrain path.

Environment (Linux / Windows): 
Locally with unit-tests, e2e run by github workflow.

Verified the fix of #1277 by actually launching +100 instances and interrupting them simultaneously (tested both reproduction of the bug and the fix of the bug).

Kubernetes Version: 
1.32

### DIFF
--- a/pkg/interruptionevent/draincordon/handler.go
+++ b/pkg/interruptionevent/draincordon/handler.go
@@ -81,7 +81,18 @@ func (h *Handler) HandleEvent(drainEvent *monitor.InterruptionEvent) error {
 	}
 
 	if drainEvent.PreDrainTask != nil {
-		h.commonHandler.RunPreDrainTask(nodeName, drainEvent)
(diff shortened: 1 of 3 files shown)


Lu-David:
Since you were able to reproduce the original bug, can you also test that this PR fixes the bug? Unit tests and e2e tests won't be enough

Lu-David (approved):

